PR #2, opened 2017-10-27
Update

aburrell:
Merge updates from rstoneback/master


aburrell:
Did this the wrong way around :(

PR #8, opened 2018-03-01
Added support for ISS FPMU

jklenzing:
Added support for the Floating Potential Measurement Unit (FPMU) on board the International Space Station (ISS) through CDAWeb.  Currently no cleaning routine.


aburrell:
I need write access to work on this pull request :-(

aburrell:
Ok, I found where the tests are, and it looks like this should be tested (since there's an exclusion list that removes anything that can't be tested using the standard tests and this isn't on it).  However, running python 2.7 and 3.4, I get 8 testing failures that I'm not sure how to approach.  I think your attention is needed on this, @rstoneback

PR #17, opened 2018-03-24
ICON EUV Initial Support

rstoneback:
Added experimental support for ICON EUV level-2 data. The preamble on variable names is automatically removed for brevity. Unit tests are included. A note was added to the change log. The instrument was added to the documentation.


rstoneback:
Reduction in test coverage is a result of not being able to download and load a test day of ICON EUV data.

PR #23, opened 2018-04-02
Fix platform dependent strftime

mddepew:
Modify strftime to use %m/%d/%Y instead of %D (not available on all
platforms)


rstoneback:
Thanks Matthew. Can you merge into develop in the future?

Git also classifies all the lines in the file as being changed. Do you know why?

PR #27, opened 2018-04-11
Meta docstrings

rstoneback:
Updated notes and examples for Meta class to include some info about the base set of metadata attributes tracked by Meta.

Added a note about using these values for programatic access to metadata as well as examples for setting these values.

Added a docstring to the special equal function.

Removed redundant variable use in the hidden attribute setter function.

### DIFF
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -42,9 +42,7 @@ class Meta(object):
     fill_label : str
         String used to label fill value in storage. Defaults to 'fill' per
         netCDF4 standard
-    
-        
-        
+
     Attributes
     ----------
     data : pandas.DataFrame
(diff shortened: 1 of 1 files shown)


rstoneback:
Good points. Will do.

On Apr 11, 2018, at 9:53 AM, Matthew Depew <notifications@github.com<mailto:notifications@github.com>> wrote:


@mddepew approved this pull request.

Looks good to me.
In _label_setter, value and actual are kind of ambiguous. current_label and new_label might be more descriptive names for what is happening. It looks like you're not calling them as named parameters anywhere, so it shouldn't be an issue to change the names. Should still be good to merge though.

aburrell:
I agree with @mddepew .  Looks good to merge after clarifying changes.

rstoneback:
I updated the variable names as suggested. Improved nearby function variable names.

PR #29, opened 2018-04-20
Meta translate function

mddepew:


mddepew:
Looks like there are some python3 issues. Fix to come.

rstoneback:
Thanks Matthew, 

Looks like lines 1582-1594 aren't ever being run in _instrument.py. There are strings generated by the test instrument pysat_testing2D that should trigger this code. Can you check this out?

mddepew:
Those lines are not exercised by the pysat_testing2d instrument because the string columns are higher-order columns that are members of a dataframe.
Those lines are for exporting a column of strings that are a part of the top level dataframe. There are strings the pysat_testing instrument that would meet this criterion, but it doesn't look like that instrument is currently used in testing. I can add a test that would exercise those lines if you want.

rstoneback:
I see a few tests in in my version of test_utils.py with pysat_testing. One example function name, test_basic_writing_and_reading_netcdf4_default_format, line 95.


On Apr 23, 2018, at 10:51 AM, Matthew Depew <notifications@github.com<mailto:notifications@github.com>> wrote:


Those lines are not exercised by the pysat_testing2d instrument because the string columns are higher-order columns that are members of a dataframe.
Those lines are for exporting a column of strings that are a part of the top level dataframe. There are strings the pysat_testing instrument that would meet this criterion, but it doesn't look like that instrument is currently used in testing. I can add a test that would exercise those lines if you want.

mddepew:
Ah, I see that now, grep missed it because it was called by module rather than by string. The issue is that there is no metadata for those variables, so it's throwing a key error exception and not reaching those lines. I can modify the pysat testing instrument to provide metadata for those variables and the export will no longer exit before those lines. What should the behavior be when trying to export a variable that does not exist in the meta object?

rstoneback:
Ah, ok. I think I was under the mistaken impression those lines used to be covered and weren’t now.
I’d appreciate it if you could add metadata for one the string columns. Might also be good to keep a lack of meta information on at least one column so we can test how we react.

Previously I had the code set up to print a statement to console there is no metadata and move on.

Ideally there should always be an entry in the metadata for all the variables. There is none in this case because of how I put together the simulated data.
Since users can currently replace the .data attached to an instrument object without our knowledge we can’t currently ensure all variables in the data are represented in the Meta object.

A potential cheap fix in the file writing routine would be to compare variables in .data and those in the metaobject. If they aren’t in the meta, do an empty assignment on that metadata variable name and it will produce a set of default metadata values.

The longer term fix is to make .data an attribute on the Instrument object so we know when and if things change and enforce consistency between .data variables and the meta object.



On Apr 23, 2018, at 11:11 AM, Matthew Depew <notifications@github.com<mailto:notifications@github.com>> wrote:


Ah, I see that now, grep missed it because it was called by module rather than by string. The issue is that there is no metadata for those variables, so it's throwing a key error exception and not reaching those lines. I can modify the pysat testing instrument to provide metadata for those variables and the export will no longer exit before those lines. What should the behavior be when trying to export a variable that does not exist in the meta object?

PR #37, opened 2018-05-09
Meta distillation

rstoneback:
Cleaned up the expansion of features in Meta. Runtime for meta test suite reduced from 10 seconds to less than 3. The __setitem__ method was simplified for both meta and instrument. Operation is tighter overall. When dealing with lower dimensional metadata, replaced returning an empty Meta instance in 'children' with None. Removed functions that weren't needed. Added a keep, drop, and merge methods that the Instrument objet is going to hook into.

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -381,72 +381,45 @@ class Instrument(object):
         """
         
         # add data to main pandas.DataFrame, depending upon the input
-        
-        if isinstance(new, dict):
-            # input dict must have data in 'data', the rest
-            # is presumed to be metadata
-            # metadata should be included in dict
-            in_data = new.pop('data')
-            if isinstance(in_data, pds.DataFrame):
-                # input is a dataframe
-                pass
- 
-            elif isinstance(in_data[0], pds.DataFrame):
-                # input is a list_like of frames
-                # this is higher order data
-                # this process ensures
-                if key not in self.meta.keys_nD():
-                    ho_meta = _meta.Meta()
-                    for ho_key in in_data[0]:
-                        ho_meta[ho_key] = {}
-                    self.meta[key] = ho_meta
+        # aka slice, and a name
+        if isinstance(key, tuple):
+            self.data.ix[key[0], key[1]] = new
+            self.meta[key[1]] = {}
+            return 
+        elif not isinstance(new, dict):
+            # make it a dict to simplify downstream processing
+            new = {'data': new}
             
-            # assign data and any extra metadata
-            self.data[key] = in_data
-            self.meta[key] = new
+        # input dict must have data in 'data', 
+        # the rest of the keys are presumed to be metadata
+        in_data = new.pop('data')
+        if isinstance(in_data, pds.DataFrame):
+            # input is a dataframe
+            # multiple 1D datasets, no need for more
+            # provides a good filter for the next step of if chain
+            pass
+        elif isinstance(in_data[0], pds.DataFrame):
+            # input is a list_like of frames
+            # this is higher order data
+            # this process ensures
+            if ('meta' not in new) and (key not in self.meta.keys_nD()):
+                # create an empty Meta instance but with variable names
+                # this will ensure the correct defaults for all subvariables
+                # meta can filter out empty metadata as needed, the check above reduces
+                # the need to create Meta instances
+                ho_meta = _meta.Meta(units_label=self.units_label, name_label=self.name_label,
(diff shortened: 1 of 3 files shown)


pysat/_instrument.py
```diff
@@ -381,72 +381,45 @@ def __setitem__(self, key, new):
...
-                raise ValueError("No support for supplied input key")
+        # input dict must have data in 'data', 
+        # the rest of the keys are presumed to be metadata
+        in_data = new.pop('data')
+        if isinstance(in_data, pds.DataFrame):
+            # input is a dataframe
+            # multiple 1D datasets, no need for more
+            # provides a good filter for the next step of if chain
+            pass
+        elif isinstance(in_data[0], pds.DataFrame):
```
mddepew: This will throw an exception if in_data is not iterable. If non-iterables are not allowed, you should throw an exception. If they are allowed, you should test for the presence of \_\_iter\_\_ on in_data before executing this line.

pysat/_meta.py
```diff
@@ -478,7 +482,12 @@ def __setitem__(self, names, input_data):
 
         elif isinstance(input_data, Meta):
             # dealing with higher order data set
-            
+            # names is only a single name here (by choice for support)
+            if (names in self._ho_data) and (input_data.empty):
```
mddepew: So, if you assign an empty meta object to a meta object label that already has higher order metadata, nothing happens? Is there a case where the user might want to replace an existing ho meta object with a new empty one?

pysat/_meta.py
```diff
@@ -389,177 +390,113 @@ def __str__(self, recurse=True):
...
+                new_name = self.attr_case_name(name)
+                if new_name != name:
+                    input_data[new_name] = input_data.pop(name)
+
+            # time to actually add the metadata           
+            for key in input_data:
+                if key not in ['children', 'meta']:
+                    for i, name in enumerate(names):
+                        to_be_set = input_data[key][i]
+                        if hasattr(to_be_set, '__iter__') and not isinstance(to_be_set, basestring):
```
mddepew: Minor point, but basestring doesn't have \_\_iter\_\_, so if you want to check for an iterable collection, you don't have to explicitly exclude strings.

rstoneback: Good to know. One of the iterable checks I've tried before did work on strings and lists/arrays/etc.

mddepew: That is really strange that a test is failing. Is it failing on 2.7? The snippets below are from python2.7
>>> a = 'test'
>>> hasattr(a,'__iter__')
False
>>> hasattr(a,'__iter__')
False
>>> a = u'test'
>>> hasattr(a, '__iter__')
False
>>> a = [1,2,3]
>>> hasattr(a, '__iter__')
True

rstoneback:
No, it is ok on 2.7 . Tested this one locally first :)

Python 3 had the issue and a bunch of errors. I. put the line back in.

Oh well. I have to say I was really excited by being able to distinguish between strings and an iterable easily.


On May 9, 2018, at 1:34 PM, Matthew Depew <notifications@github.com<mailto:notifications@github.com>> wrote:


@mddepew commented on this pull request.

In pysat/_meta.py<https://github.com/rstoneback/pysat/pull/37#discussion_r187135544>:

> -                                if isinstance(to_be_set[0], basestring):
-                                    self.data.loc[new_item_name, item_key] = '\n\n'.join(to_be_set)
-                                else:
-                                    warnings.warn(' '.join(('Array elements are disallowed in meta.',
-                                                  'Dropping input :', item_key)))
+            for name in keys:
+                new_name = self.attr_case_name(name)
+                if new_name != name:
+                    input_data[new_name] = input_data.pop(name)
+
+            # time to actually add the metadata
+            for key in input_data:
+                if key not in ['children', 'meta']:
+                    for i, name in enumerate(names):
+                        to_be_set = input_data[key][i]
+                        if hasattr(to_be_set, '__iter__') and not isinstance(to_be_set, basestring):


That is really strange that a test is failing. Is it failing on 2.7? The snippets below are from python2.7

a = 'test'
hasattr(a,'iter')
False
hasattr(a,'iter')
False
a = u'test'
hasattr(a, 'iter')
False
a = [1,2,3]
hasattr(a, 'iter')
True

pysat/_meta.py
mddepew: Yep, just checked. str has \_\_iter\_\_ in python3. Oh well, there goes that handy trick.

PR #38, opened 2018-06-12
SGP4 and PyEphem

rstoneback:
Added a new Instrument to pysat that uses a combination of SGP4 and PyEphem to support the propagation of satellite orbit data forward or backward in time. Orbit information is provided using Two Line Elements (TLEs). The spacecraft location is then used as input to IRI, HWM, ApexPy, and IGRF to provide a simulated atmosphere, geomagnetic field, and magnetic coords (mlat, mlong, mlt).

To support this code I also added some basic coordinate conversion in pysat.coords, along with supporting unit tests. Coordinate conversion currently only supports Earth Centered Earth Fixed (ECEF), geocentric (spherical Earth), and geodetic (WGS84).

While developing this code I ran into a bug with file iteration. Fix included.

Additional development is required for full simulation support.  I'm making the pull request now to get this on the radar, see if I'm getting what I need, and hopefully into develop by CEDAR. 

Everything works as is, but the plan for the future:
1) Add conversion from East/North/Down coordinates to an ECEF basis
 - IGRF outputs geomagnetic field in END, need to get it into ECEF like others
2) Create routine to project vectors in ECEF onto spacecraft frame. Code is already in the HWM support but it should be pulled out into a separate routine.
3) Potentially add ion drift simulations. Meridional ion drifts are in Fejer model. To add this I need to bring in my full magnetic coordinate system for ion drifts. Zonal and field aligned drifts will have to be coarsely simulated using simple functions. Simulation is probably too strong a word. An ion drift signal will be created that is useful for testing instrument fitting performance but shouldn't be confused with a model of the ion drifts.


rstoneback:
Future features 1 and 2 above have been added. Pyglow doesn't have Fejer model so ion drifts are still pending.

aburrell:
It'd be nice to add AACGMV2 as well, to have a fuller range of magnetic coordinates.  Could easily be beyond the scope of this pull request, though.

pysat/_instrument.py
```diff
@@ -1122,7 +1122,9 @@ def next(self, verifyPad=False):
         if self._iter_type == 'date':
             if self.date is not None:
                 idx, = np.where(self._iter_list == self.date)
-                if (len(idx) == 0) | (idx+1 >= len(self._iter_list)):
+                if len(idx) == 0:
+                    raise StopIteration('Outside the set date boundaries.')
+                if idx[-1]+1 >= len(self._iter_list):
```
aburrell: Should these two if statements be combined?  The error message is the same.  If not, I recommend making the error messages slightly different.

pysat/_instrument.py
```diff
@@ -1158,7 +1160,9 @@ def prev(self, verifyPad=False):
         if self._iter_type == 'date':
             if self.date is not None:
                 idx, = np.where(self._iter_list == self.date)
-                if (len(idx) == 0) | (idx-1 < 0):
+                if len(idx) == 0:
+                    raise StopIteration('Outside the set date boundaries.')
+                elif idx[0] == 0:
```
aburrell: Same as previous comment.

pysat/coords.py
```diff
@@ -0,0 +1,263 @@
...
+        Height (km) above presumed spherical Earth with radius 6371 km.
+        
+    Returns
+    -------
+    x, y, z
+        numpy arrays of x, y, z locations in km
+        
+    """
+
+    r = 6371. + altitude
```
aburrell: Might be more elegant to define the earth radius as a variable at the top and give a reason for choosing this particular definition of the radius.

pysat/coords.py
```diff
@@ -0,0 +1,263 @@
...
+        Geodetic Height (km) above WGS84 reference ellipsoid.
+        
+    Returns
+    -------
+    x, y, z
+        numpy arrays of x, y, z locations in km
+        
+    """
+
+    b = 6356.75231424518
```
aburrell: Comments stating what a and b are would be good.  Also, could have these defined up top?

pysat/coords.py
```diff
@@ -0,0 +1,263 @@
...
+    -------
+    latitude, longitude, altitude
+        numpy arrays of locations in degrees, degrees, and km
+        
+    """
+    
+    method = method or 'closed'
+    # calculate satellite position in ECEF coordinates
+    b = 6356.75231424518
+    a = 6378.1370
```
aburrell: Same as last comment.

pysat/instruments/pysat_sgp4.py
```diff
@@ -0,0 +1,815 @@
...
+    -------
+    None
+        Modifies pysat.Instrument object in place to include ECEF velocity 
+        using naming scheme velocity_ecef_* (*=x,y,z)
+        
+    """
+    
+    x = inst['position_ecef_x']
+    vel_x = x.values[2:] - x.values[0:-2]
+    vel_x /= 2.
```
aburrell: Is there a reason not to do:

vel_x = (x.values[2:] - x.values[0:-2]) / 2.0

?

pysat/instruments/pysat_sgp4.py
```diff
@@ -0,0 +1,815 @@
...
+    inst['unit_zonal_wind_ecef_z'] = 0*inst['position_ecef_x']
+    
+    # calculate meridional unit vector in ECEF
+    # meridional wind: north - south; positive north
+    # mer direction completes RHS of position and zonal vector
+    mag = np.sqrt(inst['position_ecef_x']**2 + inst['position_ecef_y']**2 + inst['position_ecef_z']**2)
+    unit_pos_x = inst['position_ecef_x']/mag
+    unit_pos_y = inst['position_ecef_y']/mag
+    unit_pos_z = inst['position_ecef_z']/mag
+
```
aburrell: Seems like there are a lot of places where unit vectors are needed.  Would probably be worthwhile to write a little unit_vector routine.

pysat/_instrument.py
rstoneback: If idx=[] then idx[-1] produces an error so both checks can't be in the same line. I switched line 1127 to an elif to prevent said error, though the raise prevents it anyway. I updated the error text for the first case to be more meaningful.

pysat/_instrument.py
rstoneback: Same

pysat/coords.py
rstoneback: Done.

pysat/instruments/pysat_sgp4.py
rstoneback: Nope. Fixed.

pysat/instruments/pysat_sgp4.py
rstoneback: Agreed. normalization and cross product support added to coords.py

rstoneback:
Tests are timing out on Travis CI due to some tests in the constellation class. I've reduced the test lengths but the change is in the xarray pull.

pysat/coords.py
rstoneback: same

rstoneback:
All the changes are in, including AACGMv2, and updates to the documentation.

rstoneback:
Between pyglow and pysatCDF the log gets long enough that travis CI cuts us off. pysatCDF generates a bunch of tab warnings from NASA source. pyglow generates a ton of warnings also because of old Fortran code. I thought that using pip to install pysatCDF covered up a bunch of those warnings but that doesn't appear to be the case. I don't see an option on Travis to increase the log length limit. I can remove pyglow from the build but we won't be able to test sgp4 in general.

PR #45, opened 2018-07-04
DMSP IVM support via Madrigal

rstoneback:
Added DMSP_IVM support. Improved output from test_instruments. CDAWebmethods updated to support both sat_id and tag. Documentation updated.


rstoneback:
Self approved to meet deadline

PR #66, opened 2018-08-16
Add uneven bins to median2D in ssnl.avg.py

JonathonMSmith:
The only change I'm suggesting is the added capability of making the bins for median2D capable of being uneven. 
The way I have structured this is that the user may input a sequence into bin1/2 that is taken as all of the bin edges, or they may specify the number of bins with bin1/2 and a range with range 1/2.
Perhaps instead of asking if the user has specified a range, it should ask if bin1/2 is a number or sequence.
Also will probably need some error handling for input type mismatches... if this change seems like something worth adding.

There are other changes in this fork, but I'm NOT suggesting changing median to nanmedian. It's something I like, but if introduced would have to be introduced as a flag so the user can choose and I haven't done this.

### DIFF
--- a/pysat/ssnl/avg.py
+++ b/pysat/ssnl/avg.py
@@ -12,7 +12,7 @@ import pandas as pds
 import collections
 
 def median2D(const, bin1, label1, bin2, label2, data_label, 
-             returnData=False):
+             returnData=False, auto_bin=True):
     """Return a 2D average of data_label over a season and label1, label2.
 
     Parameters
@@ -46,8 +46,12 @@ def median2D(const, bin1, label1, bin2, label2, data_label,
 
     # create bins
     #// seems to create the boundaries used for sorting into bins
-    binx = np.linspace(bin1[0], bin1[1], bin1[2]+1)
-    biny = np.linspace(bin2[0], bin2[1], bin2[2]+1)
+    if auto_bin:
+        binx = np.linspace(bin1[0], bin1[1], bin1[2]+1)
+        biny = np.linspace(bin2[0], bin2[1], bin2[2]+1)
+    else:
+        binx=bin1
+        biny=bin2
 
     #// how many bins are used
     numx = len(binx)-1
--- a/pysat/tests/test_avg.py
+++ b/pysat/tests/test_avg.py
@@ -255,3 +255,50 @@ class Test2DConstellation:
         for i, y in enumerate(dummy_y[:-1]):
             check.append(np.all(dummy_val[i, :] == y.astype(int)))
             check.append(np.all(dummy_dev[i, :] == 0))
+
+class TestSeasonalAverageUnevenBins:
+    def setup(self):
+        """Runs before every method to create a clean testing setup."""
+        self.testInst = pysat.Instrument('pysat', 'testing2D', clean_level='clean')
+
+    def teardown(self):
+        """Runs after every method to clean up previous testing."""
+        del self.testInst
+
+    def test_basic_seasonal_average(self):
+        
+        self.testInst.bounds = (pysat.datetime(2008,1,1), pysat.datetime(2008,2,1))
+        results = pysat.ssnl.avg.median2D(self.testInst, [0., 20., 167., 360.], 'longitude',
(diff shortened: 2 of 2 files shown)


aburrell:
> Perhaps instead of asking if the user has specified a range, it should ask if bin1/2 is a number or sequence.

I think this is a cleaner way to add this.

pysat/ssnl/avg.py
```diff
@@ -11,8 +11,8 @@
 import pandas as pds
 import collections
 
-def median2D(const, bin1, label1, bin2, label2, data_label, 
-             returnData=False):
+def median2D(const, bin1, label1, bin2, label2, data_label, range1=None, 
+             range2=None, returnData=False):
     """Return a 2D average of data_label over a season and label1, label2.
 
     Parameters
```
aburrell: Need to add range1/range2 to Parameters in the docstring

pysat/ssnl/avg.py
```diff
@@ -46,8 +46,15 @@ def median2D(const, bin1, label1, bin2, label2, data_label,
 
     # create bins
     #// seems to create the boundaries used for sorting into bins
-    binx = np.linspace(bin1[0], bin1[1], bin1[2]+1)
-    biny = np.linspace(bin2[0], bin2[1], bin2[2]+1)
+    if range1 is not None:
```
aburrell: Stylistically, it would be cleaner to construct the if/then using 

`if range1 is None`

aburrell:
Also, there needs to be a unit test for this that verifies uneven and even bins are working.

JonathonMSmith:
> > Perhaps instead of asking if the user has specified a range, it should ask if bin1/2 is a number or sequence.
> 
> I think this is a cleaner way to add this.

I also think this would be cleaner. However the code was previously structured such that the inputs bin1/2 were already sequences with like [l,n,m] where l is the leftmost bin edge, n is the rightmost, and m is the number of bins. I wanted to impact the structure of this file as little as possible. This way it works as it states in the documentation examples for evenly spaced bins.

JonathonMSmith:
> > Perhaps instead of asking if the user has specified a range, it should ask if bin1/2 is a number or sequence.
> 
> I think this is a cleaner way to add this.

Perhaps it would be simpler to require that bin1/2 be array_like containing bin edges. Since these values are being given to numpy's digitize function why not just take in the type of parameter that function requires. This also minimizes the modifications to the unit tests.
If this seems like a bad idea this change can continue to be modeled after matplotlib's histogram function.

rstoneback:
The current code changes the default behavior of the seasonal averaging and will break existing codes that use it. On that basis, I don't like it. I do like having the option to pass in your own bins, providing more functionality when the user desires, and maintains existing code.

My recommendation is to go back to something like the earlier version, where a flag is set, say custom_bins is True, then the seasonal averaging routine takes the input bins as is.

JonathonMSmith:
Good point. I'll make the changes.

pysat/tests/test_avg.py
```diff
@@ -255,3 +255,50 @@ def test_2D_avg(self):
         for i, y in enumerate(dummy_y[:-1]):
             check.append(np.all(dummy_val[i, :] == y.astype(int)))
             check.append(np.all(dummy_dev[i, :] == 0))
+
```
aburrell: This is a great start.  There should also be tests of cases where the uneven bins fail.

pysat/tests/test_avg.py
```diff
@@ -241,7 +241,7 @@ def test_2D_avg(self):
         for i in self.testC.instruments:
             i.bounds = (pysat.datetime(2008,1,1), pysat.datetime(2008,2,1))
         
-        results = pysat.ssnl.avg.median2D(self.testC, [0., 360., 24.], 'mlt',
+        results = pysat.ssnl.avg.median2D(self.testC, [0., 360., 24], 'mlt',
```
jklenzing: Why did 24. change to an int?

pysat/tests/test_avg.py
jklenzing: This predates this pull request, but shouldn't line 244 be 'longitude' and not 'mlt'?

jklenzing:
Unit tests to add:
Does bin input need to be int / float?  list?

PR #74, opened 2018-09-18
Develop update to model branch

aburrell:
Merge develop updates into model branch.


rstoneback:
Review Notes: I don't seem to see your model files when I click the button, so I'll do my best to fill in here. Overall, great! 

Are you thinking of replacing the xarray dataset inputs with pysat xarray instrument objects?


The only issue isn't really an issue, I just needed to get to the end. Included for completeness I suppose.
demeter_iap.py, line 214
meta = {'data type':chunk[0:10], 'data names':data_names,
            'data units':data_units}
standard label for units is 'units', 'name' or 'long_name' for the names. 'type' may or may not be ok... ok with me, the code. My bad, just got to the end of the changes. 

line 48, same file, nice

Very minor note, I think the number of - under Parameters and such is supposed to be the same number of chars as the name. That may just be sphinxdoc. Standards are mixing. Super small nit.

aburrell:
Travis fails because demeter download routine has an intentional RuntimeError at the moment.

PR #76, opened 2018-09-20
Altered order of operations in __setitem__ to increase robustness

rstoneback:


rstoneback:
Now defaults to checking for integer indexing first.

PR #80, opened 2018-09-26
Tiegcm

Gayatri012:
Hello,
I have added changes done for xarray to tiegcm branch and created a pull request on model branch.


rstoneback:
Hi Gayatri,

  I can't seem to find the instrument support file for TIEGCM. Is it included here? 

I'd like to be able to load the TIEGCM file you've been given in pysat.

Gayatri012:
> Hi Gayatri,
> 
> I can't seem to find the instrument support file for TIEGCM. Is it included here?
> 
> I'd like to be able to load the TIEGCM file you've been given in pysat.

Hi Dr. Stoneback,
I checked in my local the location (pysat.__file__) from where I'm using pysat locally and pushed files from there to tiegcm branch, then merged it with model branch. travis.ci build failed for this one too! Looks like there are a lot of changes in the _instrument file and hence preview of changes is not seen without loading the whole file. I can see that the _instrument.py file in new pull request is same as that in my local.
Let me know if you want me to merge them again.

Regards,
Gayatri

aburrell:
I think I need to do the develop to model pull for those to be incorporated.

pysat/_instrument.py
```diff
@@ -498,6 +502,7 @@ def __setitem__(self, key, new):
             self.meta[key] = new
             
         else:
+            """
```
aburrell: Are these string sections meant to maintain old code as comments?  This isn't necessary since github stores the old versions of the code.  It also makes side-by-side comparisons (such as the one I'm doing here) harder.

pysat/_instrument.py
```diff
@@ -545,6 +550,53 @@ def __setitem__(self, key, new):
...
+                # TODO: Support for encoding and Fastpath ???
+            in_data = new.pop('data')
+            # TODO: Set a common coordinates for the DataSet ? (Not individual DataArrays)
+            # This can be time or mtime, lat, lon, ilev or Z, LBC (needed?)
+            '''
+            # TODO: Add support for methods like assign_attrs(*args, **kwargs), assign_coords(**kwargs), etc.?
+            self.data[key] = xr.DataArray(data = in_data, coords = coordinates,
+                     dims = dimensions, name = name, attrs = attributes)
+            '''
+            self.data[key] = xr.DataArray(data = in_data)
```
aburrell: This is not wrong, but the style in the code is not to have white space for kwarg assignment [e.g., xr.DataArray(data=in_data)]

pysat/_instrument.py
```diff
@@ -555,6 +607,7 @@ def empty(self):
         if self.pandas_format:
             return self.data.empty
         else:
+            # TODO: What is 'time' is not one of the indexes?
```
aburrell: "What *if* 'time' is not one of the indices?" ?

pysat/instruments/pysat_testing2d_xarray.py
```diff
@@ -24,11 +24,25 @@ def init(self):
 def load(fnames, tag=None, sat_id=None):
     
     # create an artifical satellite data set
+    '''
     parts = os.path.split(fnames[0])[-1].split('-')
     yr = int(parts[0])
     month = int(parts[1])
     day = int(parts[2][0:2])
     date = pysat.datetime(yr,month,day)
+    '''
```
aburrell: @rstoneback , is it good to hardcode a year here?

pysat/instruments/pysat_testing2d_xarray.py
```diff
@@ -51,6 +65,7 @@ def load(fnames, tag=None, sat_id=None):
     mlt = np.mod(uts_root+np.arange(num)*scalar, 5820)*(24./5820.)
     data['mlt'] = (('time'), mlt)
     # do slt, 20 second offset from mlt
+    # slt - solar local time
```
aburrell: Good clarifying comment

pysat/instruments/pysat_testing2d_xarray.py
```diff
@@ -120,3 +136,17 @@ def download(date_array, tag, sat_id, data_path=None, user=None, password=None):
 alt_profile_meta['density'] = {'units':'', 'long_name':'profiles'}
 alt_profile_meta['fraction'] = {'units':'', 'long_name':'profiles'}
 meta['alt_profiles'] = {'meta':alt_profile_meta, 'units':'', 'long_name':'profiles'}
+
+
+
+data, meta = load(r'C:\Data\Gayatri\US\UTD\SORTIE\UCol Summer Internship\Data\tiegcm_icon_merg2.0_totTgcm.s_149.nc')
```
aburrell: Filename will need to be provided a different way.

pysat/_instrument.py
```diff
@@ -545,6 +550,53 @@ def __setitem__(self, key, new):
...
             # attach metadata            
             self.meta[key] = new
+
+            """
+            #print("key", key)
+            #print(new)
+                
+            if not isinstance(new, dict):
+                new = {'data': new}
+            '''
```
aburrell: Mismatched quotation mark types

aburrell:
I'll need to take another look at it, but there are some changes and clarifications that can be made now.

pysat/_instrument.py
rstoneback: That is the big question.

The current pandas format as used by pysat requires time for the main DataFrame index. In that sense, this is not anything new. If the xarray straight from the file doesn't have time as the label, it would be up to the load routine to make it so.

While we are here, should 'time' be 'epoch' instead?

I'm not trying to enforce pandas paradigms on xarray, but keeping the time assumption helps ensure the rest of pysat continues to work with little modification. The experience I have with testing the xarray pysat source is that it seemed to require that I put in 'time' specifically when I coupled it to the pysat data access and setting mechanisms. If I don't have to then great.

The trouble comes in because xarray is super general and there can be things in an xarray that are not indexed by time. I'm currently not sure what to do about this, I need a specific case to deal with. I say that based on the history of pysat... pysat is the way it is because the problems I ran into forced rethinking previously help assumptions which then generally forced evolution of the code.

pysat/instruments/pysat_testing2d_xarray.py
rstoneback: This loading should be in its own routine. The testing2d_xarray is intended as a general testing Instrument object. We need a file labeled, say ucar_tiegcm.py located in pysat/instruments. You can use the netcdf_pandas routine as a starting point. You'll need to replace the call to load_netcdf4 in the load routine there with an xarray load call. The filename to be loaded will be provided by pysat in the fnames variable. You'll also have to set pandas_format = False at the top-level. The same command is in this testing file.

You'll also need to provide a template string for how the tiegcm filenames are formatted. Normally observational files have the date and/or time in the filename. In this case, to start things off you'll need to manually update the TIEGCM name to include a date.

pysat/instruments/pysat_testing2d_xarray.py
rstoneback: The testing file generates its own fake filenames in the list_files routine. In this case I think it is ok since we have control over changing the simulated filenames. Open to discussion.

pysat/_instrument.py
rstoneback: It could very well be great to be able to have any combination of coords, dimensions, etc within the dataset. However, if we can't make any assumptions about the data we have then it becomes very difficult to do anything at all. I'd like to hold off on changes to these assumptions until I have a chance to work with xarray data in a practical context, the TIEGCM data.

pysat/instruments/pysat_testing2d_xarray.py
rstoneback: Gayatri, I'd like you to start on the individual TIEGCM instrument support file first, and then we'll iterate through the other changes you have developed once we have a better idea of where you are coming from.

pysat/_instrument.py
aburrell: So, there is probably a very good reason not to require time as the index.  However, I also think that if you want to work together with observational and modelled data then you will need to specify time.  So, in this context, it makes sense to require time as the index for modelled data.  This isn't as hard as it sounds, since you know the start time, run time, and time step of the simulations.

Also, I think the name is good.  'epoch' begs the question "which one?" whilst "time" implies datetime in a python context.

aburrell: Also, my original comment was a grammatical correction :-P

rstoneback:
That’s funny.

On Sep 28, 2018, at 8:47 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


In pysat/_instrument.py<https://github.com/rstoneback/pysat/pull/80#discussion_r221257870>:

> @@ -555,6 +607,7 @@ def empty(self):
         if self.pandas_format:
             return self.data.empty
         else:
+            # TODO: What is 'time' is not one of the indexes?


Also, my original comment was a grammatical correction :-P

pysat/_instrument.py
```diff
@@ -890,7 +890,7 @@ def _load_data(self, date=None, fid=None):
                                        date.strftime('%D')))
                 '''
                 output_str = ' '.join(('Returning', output_str, 'data for',
-                                       date.strftime('%j')))
+                                       date.strftime('%x')))
```
Gayatri012: Are we interested in date that looks like "09/29/2018" here? %j gives day of the year as a 3-digit number with zero padding if needed.

aburrell: The purpose is just to have an easily human readable date in output.  Since month/day/year is more commonly encountered than DOY, I defaulted to that when solving the merge conflicts.

rstoneback:
Given that TIEGCM instrument support has been pulled into develop via another pull, I think it is best to close this pull and restart. Work with TIEGCM and pysatMagVect should emphasize which areas of pysat are in need of attention regarding xarray/multidimensional data support.

PR #81, opened 2018-09-26
Update .travis.yml

rstoneback:
Trying to solve build failure related to pyglow. Not sure if this is ready or not but I need to get the build process in motion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,10 +48,10 @@ install:
   - ls
   # check if there is a partial download of pyglow, remove if true
   - if [[ -d "$pyglow_dir" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-      export PYGLOW_SIZE=$(du -s $pyglow_dir)
-      echo $PYGLOW_SIZE
-      echo 'PYGLOW_SIZE'
-      export CHECK_SIZE="50000"
+      export PYGLOW_SIZE=$(du -s $pyglow_dir);
+      echo $PYGLOW_SIZE;
+      echo 'PYGLOW_SIZE';
+      export CHECK_SIZE="50000";
       if [["$PYGLOW_SIZE" -lt "$CHECK_SIZE"]]; then
         rm -r ./pyglow;
         echo 'removing pyglow directory';
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ install:
   - ls
   # check if there is a partial download of pyglow, remove if true
   - if [[ -d "$pyglow_dir" && ! -e "$pyglow_dir/setup.py" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-      rmdir -r ./pyglow;
+      rm -r ./pyglow;
     fi
   # pyglow exists, install it  
   - if [[ -d "$pyglow_dir" && -e "$pyglow_dir/setup.py" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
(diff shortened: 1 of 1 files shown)


rstoneback:
The main build completed. The pull is ongoing.

.travis.yml
```diff
@@ -46,11 +46,17 @@ install:
     fi
   - pwd
   - ls
+  # check if there is a partial download of pyglow, remove if true
+  - if [[ -d "$pyglow_dir" && ! -e "$pyglow_dir/setup.py" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
```
aburrell: Is this check sufficient to ensure that all of pyglow has been downloaded?  Would it be safer to check directory size or something?

.travis.yml
```diff
@@ -46,11 +46,17 @@ install:
...
+      rm -r ./pyglow;
+    fi
+  # pyglow exists, install it  
   - if [[ -d "$pyglow_dir" && -e "$pyglow_dir/setup.py" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       cd pyglow;
       python setup.py install;
       cd ..;
     fi
+  # no pyglow, clone then run script
```
aburrell: Does this go back up to line 49 to check that the clone was successful?

rstoneback:
On Sep 26, 2018, at 2:06 PM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


In .travis.yml<https://github.com/rstoneback/pysat/pull/81#discussion_r220685649>:

> @@ -46,11 +46,17 @@ install:
     fi
   - pwd
   - ls
+  # check if there is a partial download of pyglow, remove if true
+  - if [[ -d "$pyglow_dir" && ! -e "$pyglow_dir/setup.py" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then


Is this check sufficient to ensure that all of pyglow has been downloaded? Would it be safer to check directory size or something?

I knew you were going to say something. No, this check is far from perfect. I could do a size check. Not perfect but I imagine it will be better than checking if setup.py is there. I want to be careful that simple changes to pyglow still make it through.

The best would be not having to deal with cache. Maybe money would solve our problem.....

In .travis.yml<https://github.com/rstoneback/pysat/pull/81#discussion_r220685996>:

>    - if [[ -d "$pyglow_dir" && -e "$pyglow_dir/setup.py" && "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       cd pyglow;
       python setup.py install;
       cd ..;
     fi
+  # no pyglow, clone then run script


Does this go back up to line 49 to check that the clone was successful?

The way it should work:
Either the clone is in the cache and all the downloads have already been done. Just need a quick install to register the package.

If it doesn’t do that step it is supposed to do the clone then it runs the associated script that does all the downloads and goes through the install process.

Due to time limits the builds that actually clone pyglow run out of time. So first pass does the clone and such. Second pass (restart testing) does the install and tests pysat.

If the clone isn’t done right then it gets stuck. Git won’t do a new clone if it evens sees a directory with the pyglow name, empty or not. Kinda like where we are now. :(

rstoneback:
I updated the travis-ci script code to check the size of the directory. It deletes the directory if the size is too small.

.travis.yml
rstoneback: If the clone isn't successful then Travis produces an error. I could add code to check for success but at the moment we don't have enough time to clone pyglow and run the pysat tests. The first round, the pyglow clone is made, downloads are performed, and then pyglow is installed. pysat testing continues but the runs out of time. The second time through, with a complete pyglow directory the script just runs the install. This leaves enough time for the pysat testing to complete.

I made a commit since this comment checking the directory size. If it is too small, I delete it all then try to clone.

PR #89, opened 2018-10-08
SGP4 plot fix

rstoneback:
Removed code that only applied if simulated IVM data was loaded.


rstoneback:
Coverage is lower since Travis CI is not surrently set up for testing plots. Once we get the time issue sorted on Travis (tests run longer than allowed 50 minutes) I can add tests for the plots.

PR #93, opened 2018-10-09
Wildcard support for nasa_cdaweb_methods.py + RBSPa MagEIS support

armandophys:
This commit was created as a necessity for querying data that would have a different file version within the same instrument. The Magnetic Electron Ion Spectrometer (MagEIS) instrument on the Van Allen Probe (RBSP) satellites is added to demonstrate the functionality. Note that wildcards must use '?' for each character rather than '*' for any character to be compatible with other routines in pysat.


rstoneback:
I made a small tweak to the download method to support filenames using both sat_id and tag. See pull #96 .

rstoneback:
Replaced by pull #97 .

Issue #83, opened 2018-09-28
Kp, F10.7, IMF forecasts


aburrell:
Extend OMNI, F10.7, and Kp routines to download forecasted values as well as observed or measured values.

aburrell:
Last 30 days of Daily Kp: https://services.swpc.noaa.gov/text/daily-geomagnetic-indices.txt
Next 3 days of Kp Forecast: https://services.swpc.noaa.gov/text/3-day-geomag-forecast.txt
Current F10.7: https://services.swpc.noaa.gov/text/27-day-outlook.txt
Next 3 days of F10.7: https://services.swpc.noaa.gov/text/3-day-solar-geomag-predictions.txt
45-day forecasted kp/ap: https://services.swpc.noaa.gov/text/45-day-ap-forecast.txt

rstoneback:
According to my meeting notes this should be on my action item list, correct? I'll work on this today unless I hear otherwise.

aburrell:
Yes, it is.

Tricky thing that came up when discussing usage with Sarah:

People may want to archive these forecasts so that future work can be replicated.  I think pysat can do this easily.  We'll just need to make sure we include the datetime that the forecast was retrieved as well as the datetime the forecast was made for.

Cheers,
Angeline

rstoneback:
Just pushed my first commits for Kp forecast support. Files are stored based upon the date of the forecast. Loading a forecast (by generation date) loads the three days of forecast Kp into pysat and a DataFrame.

aburrell:
I looked around and couldn't find any IMF forecast products, so I think it's just F10.7, kp, and ap.

rstoneback:
See pull #95

rstoneback:
I believe this is complete.

Issue #70, opened 2018-09-13
remote instrument date range


aburrell:
It would be very useful to be able to retrieve a range of dates for data that can be remotely accessed.

rstoneback:
Agreed. This feature would be part of an enhancement I submitted an issue on. Don’t have the number since I’m on the phone. Not a major deal but some work needs to be done on the Files class to make this happen.

On Sep 13, 2018, at 4:53 PM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


It would be very useful to be able to retrieve a range of dates for data that can be remotely accessed.

aburrell:
I could really use this as part of the model validation project.  Also, I think it would be useful to add that as a project on the github page.  I'll try to do that next week.

rstoneback:
Thanks for the heads up. I agree with the project listing. As you can see I haven’t quite made it there. :)

We should also talk about the ccmc meeting I had today. Nominal plan to start working with them on an ionosphere model. Perhaps tie-gcm since they have some work in that area. A guy from Michigan mentioned some GITM code you’ve written. And of course sami3 since you are currently in that area.

Too much to do.

On Sep 13, 2018, at 5:01 PM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:

I could really use this as part of the model validation project. Also, I think it would be useful to add that as a project on the github page. I'll try to do that next week.

aburrell:
Yes, I started the python analysis code for GITM.  So I'll be able to help with GITM and SAMI3.  But I support starting with TIE-GCM since that means someone else will need to do the heavy lifting :-D

rstoneback:
Closed per #128

aburrell:
@rstoneback , could you put a little example here, so we know how to do this?  I'm assuming the docs haven't been updated with an example of how to use this yet.

aburrell:
So the user will need to extract the date range from the files.  Does pysat have the file format at this point?  If so, it would be more in the spirit of this issue to provide a routine that backs out the dates from the file names, and another that then gets the first and last datetime available.

rstoneback:
No, the dates have already been extracted into date times. If you want the first and last date just grab the first and last index of the series that is returned.

On Jan 4, 2019, at 6:07 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #95, opened 2018-10-10
SW Forecast Support - Issue #83

rstoneback:
Initial code for F10.7 cm solar flux and Kp. Includes support for forecast values via the tag name, 'forecast'. F10.7 support includes 'all' data, from 1947 to the present. Perhaps Kp needs an 'all' option as well.

Added instruments to the documentation and __init__.py.


rstoneback:
Support for Kp, AP, and F10.7 forecasts. The 45-day F10.7 and Ap are in the F10.7 support, tag='45day'. Three day forecasts are in 'forecast' for both F10.7 and Kp instruments. Support also added for F10.7 'all', includes all data available from LASP.

rstoneback:
I need to tweak a line in this pull. Hold off for now.

rstoneback:
Code needs to be updated to modify test dates for forecast data. They are only going to have the current forecast. Timing needs to be robust with respect to the forecast release otherwise we will introduce a time dependent error.

rstoneback:
I also added the tweak referenced previously, but we are back to needing a tweak.

pysat/instruments/sw_f107.py
```diff
@@ -0,0 +1,342 @@
...
+platform : string
+    'sw'
+name : string
+    'f107'
+tag : string
+    '' Standard F10.7 data (single day at a time)
+    'all' All F10.7
+    'forecast' Grab forecast data from SWPC (next 3 days)
+    '45day' 45-Day Forecast data frmo the Air Force
+
```
aburrell: typo "from" not "frmo"

aburrell:
Ok, this might be a bit controversial, but it doesn't really make sense to supply a download date for the forecasts.  Can we make the download start and stop times optional, having them default to start=dt.datetime.today(), stop=start+dt.timedelta(days=1)?

rstoneback:
Sure. I don’t see why not.

The easiest way - the user would still have to put in a date but we could just ignore it.

Slightly harder but doable - I could replace the Instrument download as well, for this object only, that doesn’t require dates.

Which is weirder, a function that ignores inputs or one that has a different call structure than all the other downloads?


On Nov 9, 2018, at 8:34 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


Ok, this might be a bit controversial, but it doesn't really make sense to supply a download date for the forecasts. Can we make the download start and stop times optional, having them default to start=dt.datetime.today(), stop=start+dt.timedelta(days=1)?

aburrell:
I think if we just put default values into the instrument downloader that would do the trick.  If someone forgets to put the date in, they download today's file (or get a date error).  That way the behaviour won't change.

Also, the forecast does really ignore whatever date you put in, since it just grabs whatever is available.

aburrell:
I just ran into a problem when loading the 'recent' data.  It will download today's file if I give it the current time, but it won't load it:

```
import pysat
kp = pysat.Instrument('sw', 'kp', 'recent')
stime = dt.datetime.today()
kp.download(stime, stime+dt.timedelta(days=1))
```

> Downloading data to:  /Users/aburrell/Programs/Data/pysatData/sw/kp/recent/
> This routine can only download the current webpage, not archived forecasts
> Updating pysat file list
> pysat is searching for sw kp recent files.
> Found 1 of them.
> Updating instrument object bounds.

```
kp.files.files
``` 
> 2018-11-09    kp_recent_2018-11-09.txt
> dtype: object

`kp.load(date=stime)`

> No sw kp recent data for 11/09/18

```
stime = dt.datetime(2018,11,9)
kp.load(date=stime)
print(kp)
```

> pysat Instrument object
> -----------------------
> Platform: sw
> Name: kp
> Tag: recent
> Satellite id: 
> 
> Data Processing
> ---------------
> Cleaning Level: clean
> Data Padding: None
> Keyword Arguments Passed to load(): {}
> Custom Functions : 
>     No functions applied.
> 
> Orbit Settings
> --------------
> Orbit properties not set.
> 
> Local File Statistics
> ---------------------
> Number of files: 1
> Date Range: 11/09/2018 --- 11/09/2018
> 
> Loaded Data Statistics
> ----------------------
> Date: 11/09/2018
> DOY: 313
> Time range: 10/11/2018 00:00:00 --- 11/09/2018 21:00:00
> Number of Times: 240
> Number of variables: 3
> 
> Variable Names:
> Kp                            high_lat_Kp                   mid_lat_Kp

rstoneback:
That’s a bit strange. I wonder if the datetimes have a different timezone setting or something….
That or there is a bug in basic file/date selection. Perhaps something related to the file list only having one element. I don’t think that specific case is in the Files class testing routines.
Thanks for the notice.

jklenzing:
> I think if we just put default values into the instrument downloader that would do the trick. If someone forgets to put the date in, they download today's file (or get a date error). That way the behaviour won't change. Also, the forecast does really ignore whatever date you put in, since it just grabs whatever is available.

Having a default date for the instrument object is useful.  For the rockets, there is only really one date per mission, so this would be helpful for future improvements.

rstoneback:
Ok. I will sort out a proper automatic download date for the forecast and it’ll probably provide some guidance of a more systematic path forward for the rocket data.

Cheers.

On Nov 13, 2018, at 8:49 AM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


I think if we just put default values into the instrument downloader that would do the trick. If someone forgets to put the date in, they download today's file (or get a date error). That way the behaviour won't change. Also, the forecast does really ignore whatever date you put in, since it just grabs whatever is available.

Having a default date for the instrument object is useful. For the rockets, there is only really one date per mission, so this would be helpful for future improvements.

rstoneback:
Let me start with the date thing first.

datetime.today() includes the current time. When pysat receives a date to load, it uses that via slicing on files list to get the correct range of files. If the time (hour, min, sec) is included, it messes that up. Since datetimes are accepted in multiple places, we should probably implement two functions, 
1. A function that strips hour, minute, second, etc from an input datetime and returns one with only year, month, day. We should probably apply this to all methods that accept a datetime.
2. A today helper function attached to the instrument object that gives year, month, day only.

rstoneback:
If I created a today method and a tomorrow helper method I could then use those as defaults for the download method that would make some sense, and it would also sort out the request for default download of the latest data.

rstoneback:
I just pushed a commit that should address the feedback.

Download dates are now optional. Defaults to yesterday to tomorrow. 
Today, yesterday, and tomorrow are now part of the Instrument object. Could be put in utils I suppose.
The dates are filtered on load and download method. Angeline's Kp example now works for me.

To maintain the ability to load previous forecasts and ensure you are loading the latest forecast, I repeat the most recent file with tomorrow's date. Loading the forecast data can be ensured via

    kp = pysat.Instrument('sw', 'kp', tag='recent')
    kp.download()
    kp.load(date=kp.tomorrow())

The best way forward for the rocket data is to ignore the date inputs in the rocket download routine and only download the relevant data, just like the forecast routines.

aburrell:
> The best way forward for the rocket data is to ignore the date inputs in the rocket download routine and only download the relevant data, just like the forecast routines.

Also, when we get the instrument ranges sorted (see #70 ), the instrument download/load routines can check to ensure the input date is within the specified limits and even supply an instrument-specific default (e.g., times from the first instrument paper).

Issue #94, opened 2018-10-09
Documentation lacks information on sat_id functionality and use.


armandophys:
Some missions use multiple satellites and require an ID to differentiate. Currently, there is no documentation on how to do this properly.

rstoneback:
See pull #96 for a partial fix

jklenzing:
Updates to text here:
https://github.com/rstoneback/pysat/blob/documentation/docs/new_instrument.rst

jklenzing:
Closed with merge of #175.

PR #96, opened 2018-10-11
Added template instruments, related to Issues #94

rstoneback:
Trying to make adding new instruments easier.

Added a template for a general instrument, includes code for both xarray and pandas.
Uses pysat functions for list_files.

Added a template for using NASA CDAWeb methods. I expanded the download routine to include support for both sat_id and tag. Previously it only included tag. I updated associated instrument modules using CDAWeb methods.

Updated the supported instruments documentation. Trying a new section just for templates.

There is overlap between the template_instrument and the pre-existing netcdf4_pandas method. There is some distinction, the netcdf4 instrument can actually be used to load files. The template instrument docstrings are a bit more to the point. Perhaps a better basis to start a new instrument. I actually forgot about netcdf4_pandas until after I wrote this new template.

### DIFF
--- a/pysat/instruments/template_cdaweb_instrument.py
+++ b/pysat/instruments/template_cdaweb_instrument.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """This is a template for a pysat.Instrument support file that
-utilizes CDAWeb methods. Modify this file as needed when adding a 
+utilizes CDAWeb methods. Copy and modify this file as needed when adding a 
 new Instrument to pysat.
 
 This is a good area to introduce the instrument, provide background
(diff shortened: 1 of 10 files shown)


pysat/instruments/template_cdaweb_instrument.py
```diff
@@ -0,0 +1,177 @@
+# -*- coding: utf-8 -*-
+"""This is a template for a pysat.Instrument support file that
+utilizes CDAWeb methods. Modify this file as needed when adding a 
+new Instrument to pysat.
+
```
aburrell: "Modify this file..." -> "Copy and modify this file..."

pysat/instruments/template_instrument.py
```diff
@@ -0,0 +1,339 @@
...
+    # remove these variables from xarray
+    data = data.drop(['p0', 'p0_model', 'grav', 'mag', 'timestep'])
+            
+    return data, meta
+
+
+def list_files(tag=None, sat_id=None, data_path=None, format_str=None):
+    """Produce a list of files corresponding to PLATFORM/NAME.
+
+    This routine is invoked by pysat and is not intended for direct
```
aburrell: Should these extra comments be included in the bottom in a section called "Notes"?

rstoneback:
Review comments were incorporated.

PR #97, opened 2018-10-12
Mageis Full tag/sat_id support and update to sat_id issue

armandophys:
I added the same changes as issue #93 to support sat_id tags. But, I had issues with downloading cnofs_ivm data with this change so I added an "if" statement in nasa_cdaweb_methods to handle the tags correctly.

The MagEIS instrument now has full support to download files based on the sat_id and tag (cleaning level in the case of RBSP). I used the package "collection.defaultdict" to create a double "for" loop to generate all the necessary tags. This could be used a template for multi-satellite missions with different cleaning levels to generate the necessary tags without making the instrument file appear too cluttered.


rstoneback:
RBSP files include 3D data, pysatCDF requires some modification before this works. This pull can't be completed until that is sorted out.

rstoneback:
I'm going to close this since my availability for pysatCDF hasn't been very good. I'll ping you when things look different.

Issue #121, opened 2018-11-15
JRO ISR files span more than one day


aburrell:
Despite providing a file for each data, the Madrigal JRO ISR data within a file may span multiple days.  This is problematic when trying to do any of the longer term analysis things in pysat.  Not sure what the most pysatic way to fix this is.

rstoneback:
Thanks for the heads up.

That is a weird practice from Madrigal.

On Nov 15, 2018, at 10:10 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


Assigned #121<https://github.com/rstoneback/pysat/issues/121> to @rstoneback<https://github.com/rstoneback>.

aburrell:
I suggest adding a date check to the madrigal_methods.load routine.  It could take "date" from the _instrument.load routine as a keyword and remove any data that falls outside of that date.  Thoughts @rstoneback ?

rstoneback:
That is better than what I was originally thinking. Is the data that will be parsed out the same as if we requested that day from Madrigal?

aburrell:
If you request the day from Madrigal you get the same file with the same extended day range.  If they change something and only give us exactly the requested day, this addition wouldn't break the file reading.

rstoneback:
You answered things as I updated my comment. Load filtering sounds good.

aburrell:
:-P . I was too quick for you!  Let's go with load filtering.

rstoneback:
That is the truth!

On Nov 23, 2018, at 6:08 PM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


:-P . I was too quick for you! Let's go with load filtering.

rstoneback:
Fix is in pull #127. Pending review and verification. The JRO data I have doesn't appear to have this problem. I only have a couple days though.

rstoneback:
Fixed in pull #127

PR #127, opened 2018-11-28
Fix for issue #121

rstoneback:
Downselects loaded jro data to correspond with supplied date.


rstoneback:
I can certainly add it to madrigal methods. I don't see that there is any downside. Back in a few.

rstoneback:
Changes are in

rstoneback:
The pull passes all tests in python 2 but the push, same code, isn't passing the test because of pyglow. The directory is small, and is being deleted (or the command is being issued), but git clone fails every time. I'm going to merge as this is a pyglow and travis issue, not a pysat one.

PR #136, opened 2019-01-04
Issue #70

rstoneback:
Added remote_date_range function that returns first and last date.

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -1239,6 +1239,19 @@ class Instrument(object):
         
         return self._list_remote_rtn(self.tag, self.sat_id)
 
+    def remote_date_range(self):
+        """Returns fist and last date for remote data.
+        
+        Returns
+        -------
+        List
+            First and last datetimes obtained from remote_file_list
+        
+        """
+        
+        files = self.remote_file_list()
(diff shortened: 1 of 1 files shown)


rstoneback:
Great! After my last email I realized it was 1) not intuitive that going for dates meant looking at remote files unless the user is me
And 2) even leaving a couple lines to the user for a common desire isn’t ‘easy to use’.

Thanks for the reminder.

Cheers.


I’m crossing streams but the work on pydineof is going well and is making me feel better about getting a validated setup. I’m currently making a method to go from a normal numpy array to a format suitable for dineofs. Makes things easier and significantly reduces the surface area for errors, particularly for LoS drift vectors.

It will mean going back to modify existing functions to use the new methods. It should mostly be deleting code and replacing it with a method call though.

I clicked the button for review on a pull on io code. I’m more than happy to accept a full review but my intent was more of an FYI. More to come.


On Jan 4, 2019, at 12:23 PM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


@aburrell approved this pull request.

Wonderful, makes everything a bit more intuitive!

rstoneback:
I've verified that the test issues are of the CDAWeb timeout variety. Merging.

Issue #134, opened 2019-01-03
Bug in 45 day F10.7 data


aburrell:
I found a bug in the time indices for the 45 day F1.07 data.  Times are not unique.

aburrell:
Looking more closely, this error is caused by reading in only the date/data pair for the first columns in each row.

aburrell:
This bug could be identified in the future by requiring the time indices to be unique.  Now that we have xarray support, I don't see any reason why this should be the case.  Thoughts @rstoneback ?

rstoneback:
We can certainly add that condition. Seems reasonable. I’ll need to double check how SuperDARN data is loaded.

I’m in the final throws of class prep and then some meetings after. After I finish my NRL proposal I’ll review your bug fix and take a look at the unique times requirement.

Cheers,
Russell

On Jan 17, 2019, at 9:57 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


This bug could be identified in the future by requiring the time indices to be unique. Now that we have xarray support, I don't see any reason why this should be the case. Thoughts @rstoneback<https://github.com/rstoneback> ?

PR #138, opened 2019-01-17
Fixed timing bug in 45 day file

aburrell:
Fixed timing bug introduced by reading only the first date/data pair from each line in the 45-day file data blocks.

Reduced duplication by creating a function to read the 45-day file data blocks

Fixed a few obvious PEP8 violations

You can test this code by:

```
import datetime as dt
import pysat
stime = dt.datetime.today()
sw = pysat.Instrument('sw', 'f107', '45day', update_files=True)
sw.download()
sw.load(date=stime)
print sw.data
```

The date indexes should all be unique and there should be one Ap and F10.7 value for each day.

### DIFF
--- a/pysat/instruments/sw_f107.py
+++ b/pysat/instruments/sw_f107.py
@@ -194,24 +194,26 @@ def list_files(tag=None, sat_id=None, data_path=None, format_str=None):
                                        format_str=format_str)
             # pad list of files data to include most recent file under tomorrow
             if not files.empty:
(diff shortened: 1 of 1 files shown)


aburrell:
This addresses issue #134

aburrell:
Failing in the expected places.

rstoneback:
I verified test failures are related to the now standard download timeouts from CDAWeb.

Issue #124, opened 2018-11-26
utils scope


aburrell:
The scope of utils is getting too big.  I suggest we break it up into themed submodules.  For example, make utils a directory and have time, coords, stats, et cetera.

rstoneback:
See related issue #114

jklenzing:
Revisiting this as I work on unit tests for the coords.py file.  Breaking utils into four subsets, which we can refine later.  These are:
- coords (352 lines)
- stats (116 lines)
- time (286 lines)
- misc (491 lines)

The misc set basically has four functions (including the load_netcdf4, which takes up most of the space).

jklenzing:
Likewise, how should we organize the unit tests?  We could simply have a `test_utils.py` file with a different class for each submodule.

aburrell:
I propose a bit of a structural change.  Let's create a utils directory and place coords, stats, time, file, and misc (what's left there besides file utilities?) in this directory.  Then we'd be importing pysat.utils.coords, which maintains the intent behind each of these and won't bloat the top level directory as we add more functionality.

I also say we incorporate ssnl into stats.  Conceptually, I'd expect to encounter the seasonal implementations under pysat.utils.stats.ssnl.  Also, they aren't so much seasonal as long-term and a name change _might_ be appropriate.  Finally, the "invisible" routines in ssnl should be accessable in stats.  In my opinion.

However, the current testing idea says any time you have a file, it gets its own test file.  You can have classes within that file, however.  So, for example, all coords routines are tested in test_coords.  But there is one class that deals with all of the radar-related routines, another that deals with rocket-related routines, and so on.

jklenzing:
@aburrell, I've pushed my current changes to #144.  Not a working build yet, but it will be useful for structural comments.  `misc` contains `computational_form`, `set_data_dir`, `load_netcdf4`, and `scale_units`.

aburrell:
Thoughts:

* set_data_dir will need to still exist at the .utils level.  Perhaps we have a utils.py file in the utils directory and all the routines there are imported one level up.
* scale_units could be place in coords, since it allows coordinate transformation between, for example, km and m.
* @rstoneback needs to comment on computational_form, after reading the docstring I don't really know when I should use it

jklenzing:
Two more candidates for misc:
- adjust_cyclic_data
- update_longitude

I accidentally copied these over with the time functions last round.  Not sure where they should go, or if they should be their own thing (space?)

aburrell:
[...]

Issue #114, opened 2018-11-09
Coordinate transformation support


rstoneback:
Collect the coordinate transformations in pysatMagVect, along with the transformations to be added by A. Burrell in utils.py, into a pysat.coords submodule. Include outside packages into pysat's setup.py and make them available from within pysat.coords, presuming they are pip installable.

rstoneback:
pysat has seen some organization improvements that implies this coupling of MagVect and regular pysat should happen sooner than later.

jklenzing:
Since the bits of this that are breaking changes have been incorporated already (methods reorg), moving the remainder to 2.1.0.

jklenzing:
Link to #167.  What is the future of this issue with the reorganization?

rstoneback:
Sure.

PR #144, opened 2019-01-28
Coords.py

jklenzing:
Partial implementation of #114 and #124.

Moved coordinate transformations from utils.py to coords.py.

Added inverse capability to geodetic_to_geocentric transforms.

Added unit tests to check that running a transform and its inverse will produce the original inputs.


jklenzing:
@aburrell, can you look over the unit tests?  Everything in coords.py should be tested, but we may want to expand testing based on earlier versions.  Specifically:
- The 'targets' in `test_local_horizontal_to_global_geo` used the same function to generate.  Is there an external calculator for checking these?
- The other functions all use reversibility as the test, ie, does applying the inverse of the test get you back to starting point.  Could add single transforms similar to cartesian to spherical tests.
- Could add array versions of all these tests.
- Anything else I missed?

jklenzing:
Note: Leaving this here for now while we discuss #124.  Once we have a plan forward, will close and reopen a new pull request.

jklenzing:
Closing in lieu of upcoming major update to utils (see #124).

Issue #154, opened 2019-02-09
Pyglow not loading in Travis


jklenzing:
Currently, Travis CI gives an error when pyglow is loaded when testing the sgp4 files. 
```
ModuleNotFoundError: No module named 'pyglow'
```

jklenzing:
`pyglow_install.sh` no longer the default method to install `pyglow`

jklenzing:
Note for future edits: Using

```
python setup.py install --user
```

instead of

```
python setup.py install
```

will result in incorrect travis install, and `pyglow` cannot be imported from within `pysat`.

jklenzing:
Closed with merge of #155.

rstoneback:
Is there any way we can engage the --user option from the setup.py file, or something like that? I'm guessing most people aren't going to think to include the --user flag when installing pysat.

Normally Fridays are my review day but this Friday is covered up by Faculty meetings, performance reviews, proposals, and more! :)

jklenzing:
We shouldn't need the --user flag.  It's just that whatever's going on in the travis environment seemed to get annoyed if I installed pyglow with the --user option, so I removed it for pyglow.  Granted, the fact that #143 passed travis before this was merged in, so it's *possible* the issue was something they fixed up the line over the weekend.  It's ***entirely possible*** that they fixed it at the same time I implemented this change, and it has nothing to do with the changes in my code.  :)

Right now, pyglow includes --user as part of the standard install, pysat does not.  This is how my computer is set up, so I'll play with some local tests now that python 3 can run the pyglow / sgp4 code to make sure that --user doesn't affect it.

PR #143, opened 2019-01-26
cdaweb access through requests

jklenzing:
Partial fix for #79.  Replaces `ftplib` with `requests` in cdaweb_methods.download to allow access to cdaweb data via HTTPS.  Allows for future implementation of restful web services with cdaweb server.

PR #155, opened 2019-02-10
Pyglow install update / Travis streamline

jklenzing:
Fix for #154.

- Updated `.travis.yml` to install latest version of `pyglow` (python 3 compatible). 
- Streamlined `.travis.yml` and removed python version checks since not needed with `pyglow` update.
- Updated `test_instruments.py` to run sgp4 tests in python 3.6.  Should result in an increase in coverage.

When combined with final version of #143, all Travis CI tests should be passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_install:
   - pip install coveralls
   - pip install future
   - pip install matplotlib
+  - pip install netCDF4
   - pip install pysatCDF >/dev/null
   # install pyglow, space science models
   - cd ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ python:
 sudo: false
 
 cache: pip
-cache:		
-   directories:		
+cache:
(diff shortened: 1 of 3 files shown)


jklenzing:
Note for future edits:  Using `python setup.py install --user` instead of `python setup.py install` will result in incorrect travis install, and pyglow cannot be imported from within pysat.

jklenzing:
Current test failures are limited to instrument failures solved when #143 is finalized.  Confirmed that sgp4 tests successfully run in 3.6 both locally and on Travis.  This one is ready for review.

jklenzing:
Note:  Separate miniconda install on Travis removed as part of streamlining testing process.

jklenzing:
Looks like some tests were skipped due to lack of netCDF4.  Adding to Travis and testing again...

jklenzing:
And that fixed it.  @rstoneback, this one is ready.  Leaving it alone for now.

PR #162, opened 2019-02-14
Metadata update

aburrell:
Fixed bug in assigning units to metadata


pysat/model_utils.py
```diff
@@ -610,7 +610,8 @@ def extract_modelled_observations(inst=None, model=None, inst_name=[],
     # Update the instrument object and attach units to the metadata
     for mdat in interp_data.keys():
         attr_name = mdat.split("{:s}_".format(model_label))[-1]
-        inst.meta.data.units[mdat] = model.data_vars[attr_name].units
+        inst.meta.__setitem__(mdat, {inst.meta.units_label:
+                                     model.data_vars[attr_name].units})
```
jklenzing: I'm not used to seeing setitem called this way.  Shouldn't it be hidden?

aburrell: This is the only way I've found to update the metadata.

rstoneback:
Emailing here to remind myself to look at this. If you are having to call setitem then something has gone wrong.

I know I keep threatening to turn my attention back this way but it is going to happen.

On Feb 19, 2019, at 8:59 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


In pysat/model_utils.py<https://github.com/rstoneback/pysat/pull/162#discussion_r258075899>:

> @@ -610,7 +610,8 @@ def extract_modelled_observations(inst=None, model=None, inst_name=[],
     # Update the instrument object and attach units to the metadata
     for mdat in interp_data.keys():
         attr_name = mdat.split("{:s}_".format(model_label))[-1]
-        inst.meta.data.units[mdat] = model.data_vars[attr_name].units
+        inst.meta.__setitem__(mdat, {inst.meta.units_label:
+                                     model.data_vars[attr_name].units})


This is the only way I've found to update the metadata.

pysat/model_utils.py
jklenzing: Do you have a code snippet I could run to test this?

aburrell:
```
import pysat
testInst = pysat.Instrument('pysat', 'testing', '', clean_level='clean', update_files=True)
testInst.load(2009,1)
testInst.meta.__setitem__('dummy5', {testInst.meta.units_label: "Hi"})
print testInst.meta['dummy5']
```

aburrell: Oddly enough, the line I deleted works when I use it in ipython.  However, it fails in the routine.

jklenzing: OK, I think the bug is generated because you'r setting units directly rather than creating a new key in `meta.data`.

A functionally equivalent version of line 613 would be:
```
inst.meta[mdat] =  {inst.meta.units_label: model.data_vars[attr_name].units}
```

This produces the same output using the test snippet above in 3.6.

aburrell: That hasn't worked before, but I can try it again later this week.

rstoneback: What Jeff posted is how I remember it should work.

rstoneback: You can also use inst.units_label, all the labels are available from both Meta and the instrument.

ivm.meta['lst'] = {ivm.units_label : 'bloop'}

ivm.meta['lst']
Out[17]: 
units                          bloop
long_name           Local Solar Time
notes        Local Solar Time at s/c
desc                Local Solar Time
label               Local Solar Time
axis          ICON_ANCILLARY_IVM_LST
scale                         linear
value_min                          0
value_max                         24
fill                          -999.9
children                        None
Name: lst, dtype: object

rstoneback: Works in regular python as well

aburrell: OK FINE.  ;-P . Updated as suggested.  Also made some (more) PEP8 changes.

aburrell:
@rstoneback I may have been doing something similar but not quite right when attempting to not use __setitem__ before.  We should probably have some (more?) unit tests for the instruments' metadata

rstoneback:
I believe so.

Coverage for meta is pretty good at 86%, with most of tests in test_meta. Some are in test_instrument I think, since that is the main interface being tested there. Or should be.

That being said, it could use more tests. There are also a couple methods that could be reduced with a recursive call. I’ve had an internal note to go back there.

Documentation is also probably to certainly missing for a number of features added in for ICON.

Issue #160, opened 2019-02-13
.apply behaviour changing


jklenzing:
```
/home/travis/build/rstoneback/pysat/pysat/instruments/omni_hro.py:222: FutureWarning: 
Currently, 'apply' passes the values as ndarrays to the applied function. In the future, this 
will change to passing it as Series objects. You need to specify 'raw=True' to keep the 
current behaviour, and you can pass 'raw=False' to silence this warning
  center=True).apply(pysat.utils.nan_circstd, kwargs=circ_kwargs)
```

jklenzing:
From https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.window.Rolling.apply.html

Possible values for `raw`:

  - False : passes each row or column as a Series to the function.
  - True or None : the passed function will receive ndarray objects instead. If you are just applying a NumPy reduction function this will achieve much better performance.

raw=True will maintain the current behavior.  I propose this option unless we have good reason to change.  Pinging @rstoneback and @aburrell.

jklenzing:
4 results (1 commented) found in develop branch.

<img width="719" alt="screen shot 2019-02-21 at 9 40 40 pm" src="https://user-images.githubusercontent.com/19592220/53216079-60f4fe00-3621-11e9-92ef-3c803b3321c8.png">

jklenzing:
Closed with merge of #173.

Issue #158, opened 2019-02-13
Series.to_csv Future Warning


jklenzing:
```
/home/travis/build/rstoneback/pysat/pysat/_files.py:235: FutureWarning: The signature of 
`Series.to_csv` was aligned to that of `DataFrame.to_csv`, and argument 'header' will change 
its default value from False to True: please pass an explicit value to suppress this warning.
  date_format='%Y-%m-%d %H:%M:%S.%f')
```

jklenzing:
From the write up at https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html

it sounds like the new default behaviour will output column headers.  Should I add a header=False to maintain, or do we want to switch?

jklenzing:
Proposing we add `header=False` to the `Series.to_csv` statements.

9 instances found in develop branch:
<img width="602" alt="screen shot 2019-02-21 at 9 37 30 pm" src="https://user-images.githubusercontent.com/19592220/53215991-0fe50a00-3621-11e9-99b8-3047dde7d233.png">

jklenzing:
Closed with merge of #173.

PR #173, opened 2019-02-23
Future Warning

jklenzing:
Proposed fixes for #158 and #160.  Adding keywords to maintain current behavior.  Documented in corresponding issues.


jklenzing:
#158 warning only triggered on multiple calls to `_files.py`.  Adding header=False to `sw_f107.py` lines breaks functionality.   Attempting to run with only modifications to `_files.py`.

jklenzing:
3 errors due to server issues.  Restarting tests.

jklenzing:
ready for review

jklenzing:
Check into default read statements for #158

jklenzing:
Found the problem.  Using `parse_dates=True` in `pandas.read_csv` requires a header, or the first value in the file (in this case the f107 value for the test date of 2009-01-01) is treated as the data name.  Same for kp.  Latest commit defines header as True or False depending on usage to avoid this problem.

jklenzing:
Added a pep8 scrub for sw files as well.

jklenzing:
Addendum:  I think that the method of defining the pandas.Dataset used in sw_f107 triggers a default header since column names are defined.  Since pandas is changing to have `header=True` as the default, I don't think we need to define these as true, but defining them could be clearer for future us when we edit the code.  Thoughts?

pysat/instruments/sw_f107.py
```diff
@@ -381,7 +383,8 @@ def download(date_array, tag, sat_id, data_path, user=None, password=None):
                 data.iloc[idx, :] = np.nan
                 # create file
                 data.to_csv(os.path.join(data_path, 'f107_monthly_' +
-                                         date.strftime('%Y-%m') + '.txt'))
+                                         date.strftime('%Y-%m') + '.txt'),
+                            header=True)
```
rstoneback: This will impact Angeline's current F107 data set. Better moving forward but it is technically backwards incompatible.

jklenzing: I can pull out the `header=True` statements, but it looks like the `sw_f107` code in the develop branch is producing headers by default (at least on 3.6). @aburrell, can you confirm?

aburrell: All the files I know of have headers.

Issue #87, opened 2018-10-05
negative year from c/nofs instrument


JonathonMSmith:
From a fresh install and newly downloaded data in a conda virtual environment:
```
>>>ivm = pysat.Instrument(platform='cnofs', name='ivm')
>>>start = pysat.datetime(2013,12,30)
>>>stop = pysat.datetime(2014,1,30)
>>>ivm.bounds = (start, stop)
>>> ivm.load(date=stop)
```
This yields the output :
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pysat/_instrument.py", 
line 879, in load
    self.data, meta = self._load_data(date=self.date, fid=self._fid) 
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pysat/_instrument.py", 
line 622, in _load_data
    sat_id=self.sat_id, **self.kwargs)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pysat/instruments/nasa_cdaweb_methods.py", 
line 157, in load
    return cdf.to_pysat(flatten_twod=flatten_twod)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pysatCDF/_cdf.py", 
line 658, in to_pysat
    columns=new_names)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pandas/core/frame.py", 
line 379, in __init__
    copy=copy)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pandas/core/frame.py", 
line 527, in _init_ndarray
    index, columns = _get_axes(*values.shape)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pandas/core/frame.py", 
line 484, in _get_axes
    index = _ensure_index(index)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pandas/core/indexes/base.py", 
line 4974, in _ensure_index
    return Index(index_like)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pandas/core/indexes/base.py", 
line 294, in __new__
    dtype=dtype, **kwargs)
  File "~~~/anaconda3/envs/pysat_test/lib/python3.6/site-packages/pandas/core/indexes/datetimes.py", 
line 422, in __new__
[...]

PR #178, opened 2019-03-01
Check for bad datetime when loading data

jklenzing:
Uploading for comments.  Addresses #87.

If `pandas.errors.OutOfBoundsDatetime` occurs during file load, does not return data.  Functionally acts as if file is missing (ie, continues with a data gap, but returns a 'Bad datetime for...' rather than a 'No data for...' message to the screen to alert user of a problem.

Ideally, this should scan for the `xarray` equivalent as well.

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -378,17 +378,17 @@ class Instrument(object):
                     try:
                         # Assume key[0] is integer (including list or slice)
                         return self.data.loc[self.data.index[key[0]], key[1]]
-                    except KeyError:
-                        try:
-                            # Try to force as integer (eg, if ndarray)
-                            idx = self.data.index[key[0].astype(int)]
-                            return self.data.loc[idx, key[1]]
-                        except ValueError:
-                            estring = '\n'.join(("Unable to sort out data.",
-                                                 "Instrument has data : " +
-                                                 str(not self.empty),
-                                                 "Requested key : ", str(key)))
-                            raise ValueError(estring)
+                    # except KeyError:
+                    #     try:
+                    #         # Try to force as integer (eg, if ndarray)
+                    #         idx = self.data.index[key[0].astype(int)]
+                    #         return self.data.loc[idx, key[1]]
+                    except ValueError:
+                        estring = '\n'.join(("Unable to sort out data.",
+                                             "Instrument has data : " +
+                                             str(not self.empty),
+                                             "Requested key : ", str(key)))
+                        raise ValueError(estring)
             else:
                 try:
                     # integer based indexing
@@ -496,10 +496,16 @@ class Instrument(object):
                     try:
                         # Assume key[0] is integer (including list or slice)
                         self.data.loc[self.data.index[key[0]], key[1]] = new
-                    except KeyError:
-                        # Try to force conversion to integer
-                        idx = self.data.index[key[0].astype(int)]
-                        self.data.loc[idx, key[1]] = new
+                    # except KeyError:
+                    #     # Try to force conversion to integer
+                    #     idx = self.data.index[key[0].astype(int)]
+                    #     self.data.loc[idx, key[1]] = new
+                    except ValueError:
+                        estring = '\n'.join(("Unable to sort out data access.",
+                                             "Instrument has data : " +
+                                             str(not self.empty),
+                                             "Requested key : ", str(key)))
+                        raise ValueError(estring)
                 self.meta[key[1]] = {}
                 return
             elif not isinstance(new, dict):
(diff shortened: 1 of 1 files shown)


jklenzing:
Open question:  I'm not sure how we would test for this.  Building the data that would trigger the test produces a value error.

jklenzing:
Related Issue Updated:
- Removed lines in data access routines (setitem, getitem) designed to force `np.array` to integer.  `np.array` is tested for in both, and `self.data.loc[self.data.index[key[0]], key[1]]` works well with it already.  
- Leaving commented for now in case we find a bug not accounted for in unit testing.

pysat/_instrument.py
```diff
@@ -378,17 +378,17 @@ def __getitem__(self, key):
...
+                    #     try:
+                    #         # Try to force as integer (eg, if ndarray)
+                    #         idx = self.data.index[key[0].astype(int)]
+                    #         return self.data.loc[idx, key[1]]
+                    except ValueError:
+                        estring = '\n'.join(("Unable to sort out data.",
+                                             "Instrument has data : " +
+                                             str(not self.empty),
+                                             "Requested key : ", str(key)))
+                        raise ValueError(estring)
```
aburrell: Why drop the KeyError detection?  Is it no longer needed?  Why not do:

```
except KeyError or ValueError as err:
```

pysat/_instrument.py
```diff
@@ -941,14 +947,29 @@ def _load_data(self, date=None, fid=None):
...
+                data, mdata = self._load_rtn(load_fname, tag=self.tag,
+                                             sat_id=self.sat_id, **self.kwargs)
+                # ensure units and name are named consistently in new Meta
+                # object as specified by user upon Instrument instantiation
+                mdata.accept_default_labels(self)
+                bad_datetime = False
+            except pds.errors.OutOfBoundsDatetime:
+                bad_datetime = True
+                data = self._null_data.copy()
+                mdata = _meta.Meta(units_label=self.units_label,
```
aburrell: Did mdata.accept_devault_labels no longer work?

pysat/_instrument.py
jklenzing: Back when I put #125 together, this was part of the added checks since `pandas.ix` handled all of this internally, and the data access within `_instrument.py` needed to be handled in a more stringent fashion.  The intent was that the first round of checks (KeyError, TypeError) handled most cases where integers were fed into the system, while the second round (removed here) handled a numpy.array of integers.  This was overdesign on my part, as we test for this case and the first round of checks handle it.

I'm fine with leaving it in, but we should test for it and I can't think of a case where it would be needed.  Hence leaving it in as a comment in case we find a bug.

aburrell: Since we have version control and can look this bit up later if needed, I am very much opposed to leaving it in as commented code.  Given what you said I am comfortable having it removed entirely or included in a combined exception statement, as you prefer.

pysat/_instrument.py
jklenzing: If I just leave mdata.accept_default_labels, I get
```
UnboundLocalError: local variable 'mdata' referenced before assignment
```
Since the code treats finding a corrupted file the same as finding no file, mdata needs to be defined separately.  Open to suggestions for a cleaner implementation.

pysat/_instrument.py
jklenzing: ok.  removing completely.

pysat/_instrument.py
rstoneback: you could probably get away with:
mdata = _meta.Meta().accept_default_labels(self)

rstoneback:
As a general comment, I'm not sure we should trap this particular error. A bad time in the file needs to be dealt with. I think sometimes errors may need to stop things.

If we do want to trap for more errors on the load, then I think we should probably start with a more systematic process. Once we start, I think the list of trapped errors may grow.

pysat/_instrument.py
rstoneback: Yeah, if we don't have a test for it then it needs to be removed or get a test. I'm a bit late as Jeff already removed it...

rstoneback:
Some users may not be able to modify the files on the system they are using, thus if there is an error pysat needs some mechanism to deal with it. Consensus is while it is important for a bad time in a file to get fixed, in the meantime, there is data analysis to do.

pysat/_instrument.py
jklenzing: doesn't work.

aburrell: Ok, leave it as is then.

PR #164, opened 2019-02-21
Utils restructure

jklenzing:
Closes #124 and partial fix for #114.

Summary of Changes
- Moved most functions from `utils.py` into separate directories: coords, stats, time
- `set_data_dir`, `load_netcdf4`, and `computational_form` remain under utils as part of `_core.py`
- Expanded unit testing to cover most functions within the utils family
- Updated coordinate transformations using pydarn as a ref
- Occasional reworking of multi-line strings to use `''.join` notation and other pep8 polishes.

Notes
- test_units_coords.py:TestBasics.test_calc_solar_local_time (lines 81-89) passes locally but is failing on Travis CI.  **Could use an extra set of eyes here**.
- unit testing for coordinate transformations were generated using the functions here.  This will let us know if they change, but ideally, we should have a second check for the target numbers.
- `stats.median1D` remains uncovered.  Leaving while we discuss how `ssnl` fits into the new framework.
- This is the last of the major pull requests I have for now.  Everything else in my queue is instrument updates.  :D

PR #179, opened 2019-03-01
Import reload

jklenzing:
Summary of changes
- Updated import of reload statements now that python 3.3 has reached end of life (consistent with changes discussed in updates to utils #164)
- Reordered library imports for testing files as in other pull requests
- Deleted commented tests that have been moved to other testing files
- Deleted comments in setup from update of pyglow
- General pep8 scrub of changed files


pysat/tests/test_instrument.py
```diff
@@ -1,24 +1,24 @@
...
-        reload = importlib.reload
+    from importlib import reload as re_load
+else:
+    re_load = reload
 
 
 class TestBasics():
     def setup(self):
-        reload(pysat.instruments.pysat_testing)
+        re_load(pysat.instruments.pysat_testing)
```
aburrell: @rstoneback or @jklenzing , why is this necessary for a clean testing set up?

setup.py
```diff
@@ -25,11 +25,6 @@
                     'pyEphem', 'requests', 'pysatCDF', 'apexpy', 'aacgmv2',
                     'pysatMagVect', 'madrigalWeb', 'h5py', 'PyForecastTools']
 
-# keep pyglow out of requirements until pip installable
-# if sys.version_info[0] < 3:
-#     # TODO Remove when pyglow works in python 3
-#     install_requires.append('pyglow')
-
```
aburrell: Yay!

pysat/tests/test_instrument.py
jklenzing: I'm not sure.  @rstoneback?

setup.py
jklenzing: Forgot to add this to the list of changes.  :D

pysat/tests/test_instrument.py
jklenzing: Adding to agenda for today's meeting.  This has been one of my questions for a while.

setup.py
jklenzing: Wait, I think this is implemented wrong.  I think I should have just removed the if statement, not the install_requires bit.  @rstoneback?

aburrell: Ah, I think so!  I think you just remove the comments?

jklenzing: Or just add it to the list above, since we don't need to append.  Updated readthedocs truncation as well.

pysat/tests/test_instrument.py
rstoneback: I mess with the module in a couple tests, lines 210 and 222.

pysat/tests/test_instrument.py
```diff
@@ -189,19 +189,6 @@ def test_instrument_init(self):
...
-#         assert ((yr == 2009) & (doy == 1))
-#
-#     def test_getyrdoy_leap_year(self):
-#         """Test the date to year, day of year code functionality
-#         (leap_year)"""
-#         date = pds.datetime(2008,12,31)
-#         yr, doy = pysat.utils.getyrdoy(date)
-#         assert ((yr == 2008) & (doy == 366))
-
     def test_custom_instrument_load(self):
```
rstoneback: I think this method and the two below it may need to be in their own class. This whole suite of tests is called for xarray and other configs yet the modules are fixed to pysat_testing, so they aren't being run correctly for other situations. We can also move the reload to only handle these tests.

rstoneback: Ideally the tests should be modified to accept the other testing instruments. Anytime I leave a gap in testing there is sure to be a bug.

pysat/tests/test_instrument.py
```diff
@@ -559,15 +546,16 @@ def test_incorrect_creation_empty_instrument_object(self):
...
+            pass
         Dummy.name = 'help'
 
         temp = pysat.Instrument(inst_module=Dummy)
 
 
 class TestBasicsXarray(TestBasics):
     def setup(self):
-        reload(pysat.instruments.pysat_testing)
+        re_load(pysat.instruments.pysat_testing)
```
rstoneback: All of these re_loads aren't really correct due to my previous generalization that wasn't quite complete. To be consistent, it should reallybe a reload of pysat_testing_xarray, but then that module doesn't get tested properly as noted in comment above.

setup.py
rstoneback: Updated version looks good to me.

pysat/tests/test_instrument.py
aburrell: Let's put some comments above this to that effect, so that people looking for code examples will know why this is done.

pysat/tests/test_instrument.py
rstoneback: Enhancement noted by issue #182

pysat/tests/test_instrument.py
rstoneback: enhancement noted in issue #182

pysat/tests/test_instrument.py
rstoneback: That is reasonable. Issue created to remind us to more completely solve in the future.

jklenzing:
including latest updates from develop

PR #185, opened 2019-03-01
Dmsp clean bugfix

aburrell:
Fixed bug where no cleaning removed all data instead of returning all data.

### DIFF
--- a/pysat/instruments/demeter_iap.py
+++ b/pysat/instruments/demeter_iap.py
@@ -271,7 +271,7 @@ def clean(iap):
                 # Select times with at least two ion species
                 idx, = np.where(nions > 1)
     else:
-        idx = []
+        idx = slice(0, iap.index.shape[0])
 
     iap.data = iap[idx]
 
--- a/pysat/instruments/dmsp_ivm.py
+++ b/pysat/instruments/dmsp_ivm.py
@@ -243,7 +243,7 @@ def smooth_ram_drifts(ivm, rpa_flag_key=None, rpa_vel_key='ion_v_sat_for'):
     if rpa_flag_key in list(ivm.data.keys()):
         rpa_idx, = np.where(inst[rpa_flag_key] == 1)
     else:
-        rpa_idx = list()
+        rpa_idx = slice(0, ivm.index.shape[0])
 
     ivm[rpa_idx, rpa_vel_key] = ivm[rpa_idx, rpa_vel_key].rolling(15, 5).mean()
     return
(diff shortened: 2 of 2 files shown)


rstoneback:
I’m just seeing this over email. I thought the download tests ensured there was some data after loading with no cleaning applied. I’m curious why we didn’t catch this earlier

Cheers,
Russell

On Mar 1, 2019, at 5:42 PM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


@jklenzing approved this pull request.

Looks good, dmsp tests pass locally.

rstoneback:
The cleaning routine isn't called if clean_level is none.

Line 373 could be made an error.

aburrell:
@rstoneback , the only thing I can think of is that the tests don't look for all the different dmsp tags and sat ID's.

aburrell:
Do we want to do something with line 373 or merge?

jklenzing:
I was largely wondering if that line would produce the same error as above.  If so, we should fix it now.

aburrell:
It depends on the purpose of that line.  That bit is Russell's code, but reading it I think it should be replaced.  I also just checked the rest of the instruments (using the idx variable name) and found some to be fixed.

rstoneback:
Going from memory if the rpa flag isn’t in the  loaded file then that routine drops all the data.

This behavior is noted in docstring but a warning or error could be more appropriate. I can imagine someone trying to figure out why there is no data coming out of their dmsp pipeline.

On Mar 4, 2019, at 7:33 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


It depends on the purpose of that line. That bit is Russell's code, but reading it I think it should be replaced. I also just checked the rest of the instruments (using the idx variable name) and found some to be fixed.

aburrell:
Do we want the routine to drop all the data if there isn't an rpa flag, or keep it all?

rstoneback:
I believe the rpa flag is used by the routine so I think we should have a safe default. Either no data or more explicit warning.

My vote is no data.

On Mar 4, 2019, at 7:45 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:

Do we want the routine to drop all the data if there isn't an rpa flag, or keep it all?

rstoneback:
Is the empty list not working as a mechanism for dropping data?

aburrell:
It is working as a mechanism for dropping data.  That wasn't the behaviour I was expecting when using it though, so that is why there were problems.

rstoneback:
Thanks for clearing that up.

Issue #174, opened 2019-02-23
Add support to cdaweb_methods for instruments with multiple files per day


jklenzing:
Related to #58.

Some instruments on cdaweb have multiple files per day, typically one per orbit.  Download routine needs to be adjusted to handle this.

jklenzing:
Look into implementing existing routines from DMSP and COSMIC

jklenzing:
To Do:  
[ ] Update list_remote_files to allow for partial lists by year, year/month
[ ] Incorporate this into download(..., multi_file_day=True) to reduce code duplication
[ ] Improve messages to screen during download process

jklenzing:
Closed with merge of #189.

PR #189, opened 2019-03-04
cdaweb_methods enhancement -- TIMED/SABER data

jklenzing:
Addresses #174.

Summary of changes:
- Adds support to cdaweb_methods for multiple files per day.
- Adds keywords to `cdw.list_remote_files` to allow user to select which files they search for on the remote server,
- Adds support for TIMED/SABER data.

Rationale
- Some datasets have a lot of data.  SABER has produced ~15 files per day for the last 17 years.  Users can now specify year, year and month, or year/month/day to search for files.
- The year/month/day specification is tied into the download routine, so that this part of the code is now tested in Travis CI.  This also allows for easy download of files when multiple files fit the same day pattern, but other parts of the string are not identifiable.
- The list_remote_files routine currently identifies files by the extension.  Support is included to potentially identify by multiple targets, such as the tag if multiple data products stored in the same directory.  

Notes
- The code both here and develop is not compatible with searching for all files on the remote directory if multiple date-based directories are used.  A warning has been added to let the user know.

To the mesosphere!


pysat/instruments/nasa_cdaweb_methods.py
```diff
@@ -260,27 +262,58 @@ def download(supported_tags, date_array, tag, sat_id,
...
+                sys.stdout.flush()
+                remote_path = '/'.join((remote_url.strip('/'),
+                                        formatted_remote_fname))
+                req = requests.get(remote_path)
+                if req.status_code != 404:
+                    open(saved_local_fname, 'wb').write(req.content)
+                    print('Finished.')
+                else:
+                    print('File not available for ' + date.strftime('%x'))
+            except requests.exceptions.RequestException as exception:
```
aburrell: You should also output the exception message.

aburrell: Here and elsewhere

pysat/instruments/template_cdaweb_instrument.py
```diff
@@ -38,11 +38,11 @@
 from __future__ import print_function
 from __future__ import absolute_import
 
-import pandas as pds
+import functools
 import numpy as np
-import pysat
+import pandas as pds
```
aburrell: Pandas isn't a core python package, so should be after sys, but before pysat.

jklenzing: fixed.

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: fixed

.travis.yml
```diff
@@ -22,6 +22,8 @@ before_install:
   - pip install future
   - pip install matplotlib
   - pip install netCDF4
+  - pip install beautifulsoup4
+  - pip install lxml
```
rstoneback: Do we need to add these to setup.py as well?

pysat/instruments/__init__.py
```diff
@@ -11,6 +11,7 @@
            'dmsp_ivm', 'icon_ivm', 'icon_euv', 'icon_fuv', 'icon_mighti',
            'iss_fpmu',  'jro_isr', 'omni_hro', 'pysat_sgp4', 'rocsat1_ivm',
            'sport_ivm', 'superdarn_grdex', 'supermag_magnetometer',
-           'sw_dst', 'sw_kp', 'sw_f107', 'timed_see', 'ucar_tiegcm', ]
+           'sw_dst', 'sw_kp', 'sw_f107', 'timed_saber', 'timed_see',
```
rstoneback: Nice. TIMED SABER needs to be added to the official docs as well. docs/supported_instruments.rst  I think.

pysat/instruments/nasa_cdaweb_methods.py
```diff
@@ -260,27 +262,61 @@ def download(supported_tags, date_array, tag, sat_id,
...
-            if req.status_code != 404:
-                open(saved_local_fname, 'wb').write(req.content)
-                print('Finished.')
-            else:
-                print('File not available for ' + date.strftime('%x'))
-        except requests.exceptions.RequestException as exception:
-            print('File not available for ' + date.strftime('%x'))
+        if not multi_file_day:
+            try:
+                print('Attempting to download file for ' + date.strftime('%x'))
```
rstoneback: This date output should be updated to the same format used by load. Write out month name, day, year. 
date.strftime('%d %B %Y')

pysat/instruments/nasa_cdaweb_methods.py
```diff
@@ -260,27 +262,61 @@ def download(supported_tags, date_array, tag, sat_id,
...
+                                        formatted_remote_fname))
+                req = requests.get(remote_path)
+                if req.status_code != 404:
+                    open(saved_local_fname, 'wb').write(req.content)
+                    print('Finished.')
+                else:
+                    print('File not available for ' + date.strftime('%x'))
+            except requests.exceptions.RequestException as exception:
+                print(' '.join((exception, '- File not available for',
+                                date.strftime('%x'))))
```
rstoneback: same

pysat/instruments/nasa_cdaweb_methods.py
```diff
@@ -260,27 +262,61 @@ def download(supported_tags, date_array, tag, sat_id,
...
+
+        else:
+            try:
+                print('Attempting to download files for '
+                      + date.strftime('%x'))
```
rstoneback: same

pysat/instruments/nasa_cdaweb_methods.py
```diff
@@ -354,35 +412,74 @@ def list_remote_files(tag, sat_id,
...
-        for link in links:
-            if link['href'].count('/') == 1:
-                dirs.append(link['href'])
-        # only want to keep file portion of the string
-        format_str = dir_split[-1]
-    elif len(dir_split) == 2:
-        # no extra directories
-        dirs = ['']
+    if len(dir_split[0]) != 0:
+        subdirs = dir_split[0].split('/')  # Get all subdirectories
```
rstoneback: I think pep8 frowns upon comments to the side?

pysat/instruments/timed_saber.py
```diff
@@ -0,0 +1,204 @@
...
+    normally offer co-authorship to the PI, Associate PI or designated team
+    members. Co-authorship may be declined. Appropriate acknowledgement of
+    institutions, personnel, and funding agencies should be given.
+    - Users should heed the caveats of SABER team members as to the
+    interpretation and limitations of the data. SABER team members may insist
+    that such caveats be published, even if co-authorship is declined. Data
+    and model version numbers should also be specified.
+    - Pre-prints of publications and conference abstracts should be widely
+    distributed to interested parties within the mission and related projects.
+    - Note on Temperature Errors: http://saber.gats-inc.com/temp_errors.php
```
rstoneback: You may want to move this to the official Notes formatted area or even a Warning

pysat/instruments/nasa_cdaweb_methods.py
```diff
@@ -260,27 +262,61 @@ def download(supported_tags, date_array, tag, sat_id,
...
+        else:
+            try:
+                print('Attempting to download files for '
+                      + date.strftime('%x'))
+                sys.stdout.flush()
+                remote_files = list_remote_files(tag=tag, sat_id=sat_id,
+                                                 supported_tags=supported_tags,
+                                                 year=date.year,
+                                                 month=date.month,
+                                                 day=date.day)
```
rstoneback: Won't date always have a year, month, and day here? I'm a bit confused as to how the list_remote_files code for month or day is None is run in practice.

.travis.yml
jklenzing: good catch.  fixing.

pysat/instruments/__init__.py
jklenzing: added

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: fixed

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: fixed

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: fixed

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: The download routine assumes you are downloading one or more files per day.  When download calls list_remote_files, it fills in all these values.  If the user wants to just look for data in a given year, they can specify only the year, and month and day will be None automatically.

Right now, internal stuff assumes the exact date exists, but the code is flexible enough to allow the user to search for larger ranges.

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: updated

pysat/instruments/timed_saber.py
jklenzing: moved to warning.

pysat/instruments/nasa_cdaweb_methods.py
jklenzing: I think it's clearer to think of the `list_remote_files` as a method that can either be called by the user 'directly' (through remote_list_files) or indirectly through the `download` method.

Examples:
```
d0 = pysat.datetime(2009, 1, 1)
saber.download(d0, d0)
```
will download all data for 1 Jan 2009.  `download` determines which files need to be downloaded by passing through all three variables (as in the `date` statements above) to `list_remote_files`.   Since the `download` method assumes one day at a time, it always sets all three date parameters for this call.

Conversely,
```
saber.remote_file_list(year=2009, month=1)
```
is a user-invoked command that will call `list_remote_files`, which fills in missing values as None.  In this case, day=None, and all files for January are returned.

While writing this up, I realized that I probably should set up proper variables in the `remote_files_list` method instead of just passing everything through as `**kwargs` and letting cdaweb_methods sort it out.

rstoneback: OK, thanks for clearing that up.

jklenzing:
@rstoneback, two questions before I merge:
- Should I update the `remote_list_files` method to actively look for year, month, day keywords, or is using **kwargs at this level fine?
- The year, month, day keywords are only implemented in cdaweb_methods.  Should I update the new_instrument documentation?

rstoneback:
Using **kwargs is technically fine. I see pandas and matplotlib do it quite a bit. One downside is they don't always reference what methods the **kwargs are for, or how I can find out which keywords do stuff. Updating the docstring probably a decent minimum correction. The upside is **kwargs works for any changes in the future.

From a user experience point of view, I think adding the keywords may be the easiest option for folks to understand.

- Yes, updating the documentation would be great.

 - We may want to consider a FAQ type section. Whenever we have a question and it gets answered, we should put it in there.... Like, how do I add an instrument with multiple files per day? I'll post this in issues.

jklenzing:
OK.  Will update, push, and merge.  Thanks!

Issue #130, opened 2018-11-29
xarray len deprication


aburrell:
Current use of len with xarray will need to change.

~/Git/pysat/pysat/_instrument.py:938: FutureWarning: calling len() on an xarray.Dataset will change in xarray v0.11 to only include data variables, not coordinates. Call len() on the Dataset.variables property instead, like ``len(ds.variables)``, to preserve existing behavior in a forwards compatible manner.
  if len(data) > 0:

rstoneback:
I remember seeing this warning and if I remember correctly I think we are ok. We should double check though.


On Nov 29, 2018, at 10:27 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:


Current use of len with xarray will need to change in.

~/Git/pysat/pysat/_instrument.py:938: FutureWarning: calling len() on an xarray.Dataset will change in xarray v0.11 to only include data variables, not coordinates. Call len() on the Dataset.variables property instead, like len(ds.variables), to preserve existing behavior in a forwards compatible manner.
if len(data) > 0:

jklenzing:
Digging through recent Travis logs, and I don't see this FutureWarning in the list.  Is this still an issue?

rstoneback:
I still see it pop up during normal use. I am a few commits behind though. I don’t quite get the error since it doesn’t seem to do what it claims it does.

Personally, I’m only interested in getting the data variables. I’m ok with future behavior.

jklenzing:
Yeah, I've been looking at develop.  Can't seem to find more info on this issue.

rstoneback:
It pops up when using xarray based instruments. I’ve been seeing it lately when I load tiegcm data.

On Feb 23, 2019, at 8:07 AM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


Yeah, I've been looking at develop. Can't seem to find more info on this issue.

jklenzing:
Is it just line 938, or more extensive?

jklenzing:
Probably why I'm not seeing it.

rstoneback:
Here is what I get when loading TIEGCM.

In [27]: tiegcm = pysat.Instrument('ucar', 'tiegcm', file_format='tiegcm_{year:04d}_ctmt.s_{day:03d}.nc', update_files=True)
pysat is searching for ucar tiegcm files.
Found 1 of them.
Mission acknowledgements and data restrictions will be printed here when available.

[...]

PR #192, opened 2019-03-06
More testing

jklenzing:
Summary of Changes:
- Added tests for `ssnl_plot` and `_files`
- Updates to `ssnl_plot.scatterplot` for changes in matplotlib
- Reorganized `test_sw` so the Kp and F107 each have two testing classes: one with downloads and one without
- Fixed a bug where `test_utils` would reset the pysat data directory to '.' and not change it back
- Removed option to skip tests if netCDF4 does not import
- Restructured some statements to minimize skipped lines
- Tidied up unused import statements
- Reordered import statements by canonical order 
- General syntax consistency and pep8 stuff

Notes:
- Reduces run time on Travis CI by ~22 minutes on python 2.7, ~12 minutes on python 3.6.  Pushing sooner rather than later since this will aid development time.
- In general, tried to limit each test file to use either `assert_raises` or `raises`, with the exception of test_utils_coords, where it made sense to use both
- Since `._files.parse_delimited_filename` is not currently used in end-to-end testing, added test function for this until instrument is added
- Still an outstanding issue where scatterplot does not generate the second panel.  Writing up as an issue.
- Python 2.7 on Travis CI requires a display to be set for matplotlib tests to run.  Since this is minimal, setting for both 2.7 and 3.6.
- Outstanding testing improvements to be added as issues
- From local tests, test_instruments takes up about 1/3 of the runtime


jklenzing:
Pulling this back for now while I resolve some issues.  Comments welcome.

.travis.yml
```diff
@@ -41,6 +41,13 @@ install:
   # install pysat
   - cd /home/travis/build/rstoneback/pysat
   - python setup.py install
+
+# Set up a display for plot tests
```
rstoneback: This is really good. I can test my sgp4 plots now. That should help our test coverage.

pysat/ssnl/plot.py
```diff
@@ -38,64 +38,50 @@ def scatterplot(inst, labelx, labely, data_label, datalim, xlim=None,
     figs = []
     axs = []
 
-    # check if doing multiple data quantities
+    # Check for list-like behaviour of data_label
     if type(data_label) is str:
-        # max of one data type
+        data_label = [data_label]
```
rstoneback: Nice reduction.

pysat/ssnl/plot.py
```diff
@@ -38,64 +38,50 @@ def scatterplot(inst, labelx, labely, data_label, datalim, xlim=None,
...
-            ax2.set_ylim(ylim)
-            plt.hold(True)
 
     # norm method so that data may be scaled to colors appropriately
     norm = colors.Normalize(vmin=datalim[0], vmax=datalim[1])
     p = [i for i in np.arange(len(figs))]
+    q = [i for i in np.arange(len(figs))]
     for i, inst in enumerate(inst):
         for j, (fig, ax) in enumerate(zip(figs, axs)):
             if len(inst.data) > 0:
```
rstoneback: How about   if not inst.empty:   instead ?

pysat/tests/test_custom.py
```diff
@@ -151,11 +148,11 @@ def custom1(inst):
             return out
         self.add(custom1, 'add')
         self.testInst.load(2009, 1)
-        ans = (((self.testInst.data['doubleMLT'] == 2.0 *
-                 self.testInst['mlt']).all()) &
-               ((self.testInst.data['tripleMLT'] == 3.0 *
-                 self.testInst['mlt']).all()))
-        assert ans
+        ans1 = (self.testInst.data['doubleMLT'] == 2.0 *
```
rstoneback: This was a poor style I started. the ans1 and ans2 should be removed in favor of direct asserts, ans1 = should be assert instead.

pysat/tests/test_custom.py
```diff
@@ -271,7 +268,7 @@ def custom1(inst):
             out = (inst.data.mlt * 2).values
             return {'data': out, 'long_name': 'doubleMLTlong',
                     'units': 'hours1', 'name': 'doubleMLT'}
-        self.add(custom1, 'add')
+        self.testInst.custom.add(custom1, 'add')
```
rstoneback: good catches

pysat/tests/test_instruments.py
```diff
@@ -34,6 +35,18 @@ def safe_data_dir():
     return saved_path
 
 
+def remove_files(inst):
+    # remove any files
+    dir = inst.files.data_path
```
rstoneback: dir is a special word. Please avoid.

rstoneback: We may also want check in here that there aren't many files before we delete. I'd hate to somehow accidentally delete a bunch of good data.

pysat/tests/test_sw.py
```diff
@@ -159,10 +166,20 @@ def test_combine_kp_none(self):
...
 
-        combo_in = {"standard_inst": self.testInst}
+        # Load a test instrument
+        testInst = pysat.Instrument()
+        testInst.data = pds.DataFrame({'Kp': np.arange(0, 4, 1.0/3.0)},
+                                      index=[pysat.datetime(2009, 1, 1)
+                                             + pds.DateOffset(hours=3*i)
+                                             for i in range(12)])
+        testInst.meta = pysat.Meta()
+        testInst.meta.__setitem__('Kp', {testInst.meta.fill_label:
```
rstoneback: this should be 
testInst.meta['Kp'] = {testInst.meta.fill_label: np.nan}

pysat/tests/test_utils.py
```diff
@@ -36,8 +35,7 @@ def remove_files(inst):
     # remove any files
     dir = inst.files.data_path
     for the_file in os.listdir(dir):
-        if (the_file[0:13] == 'pysat_testing') & (the_file[-19:] ==
-                                                  '.pysat_testing_file'):
+        if (the_file == 'test_ncdf.nc'):
```
rstoneback: I made the name purposely long and specific to pysat to ensure we don't ever accidentally delete a user's file

pysat/tests/test_files.py
```diff
@@ -136,6 +116,35 @@ def teardown(self):
             pass
         del self.testInst
 
+    def test_parse_delimited_filename(self):
```
rstoneback: You have a few special words used by the method. While as written things should be ok, it could potentially cause some issues in the future. I'd recommend moving min to mins, dict to file_dict (?).

pysat/tests/test_files.py
```diff
@@ -52,14 +45,10 @@ def remove_files(inst=None):
 
 
 # create year doy file set
-def create_files(inst, start=None, stop=None, freq='1D', use_doy=True,
+def create_files(inst, start=pysat.datetime(2009, 1, 1),
+                 stop=pysat.datetime(2013, 12, 31), freq='1D', use_doy=True,
```
rstoneback: I believe string defaults a frowned upon. Things can get messed up there. This is the source of my preference for setting optionals to None, though I tend to use it for more than strictly required.
Default keywords should not be set to strings, lists, dicts, anything that can be modified in memory as those changes persist between calls. The method headers are only ever set once in python.

rstoneback: I started the string default, it is my bad, but now I see it.

jklenzing: Is my use of the start and stop dates ok? It still looks a little odd to me, but it does reduce the uncovered lines. Another option is to remove start and stop defaults, since these are always specified in the testing routine.

pysat/tests/test_instruments.py
jklenzing: Since we are in a temp directory, this should be fine. Maybe a check for the modified instrument file list before a delete?

pysat/tests/test_utils.py
jklenzing: Will update the test file name

pysat/ssnl/plot.py
jklenzing: That works in the unit test.  Is this a potential solution to #130 as well?

pysat/ssnl/plot.py
jklenzing: Thanks!

pysat/tests/test_custom.py
jklenzing: updating these to use separate assert statements for each statement

pysat/tests/test_sw.py
jklenzing: fixed

pysat/tests/test_files.py
jklenzing: fixed

pysat/tests/test_utils.py
jklenzing: fixed.  I think this was originally a copy and paste error from the test_files routine, where the pysat_testing?????????.pysat_testing_file routine is used.  Here the files created are just 'test_ncdf.nc'.  Updated the file name here to 'pysat_test_ncdf.nc', which should be unique.

pysat/tests/test_instruments.py
jklenzing: I think I can drop the target and just use:
```
for the_file in os.listdir(inst.files.files.values):
```
Since we've swapped directories, this list is reset after the test data is downloaded.

jklenzing: Maybe if the number of files are less than some level?  TIMED/SABER downloads 15 files as part of the test.

pysat/tests/test_files.py
jklenzing: Replaced with `freq=None`, then added a check to reset within the function.

Also, removed default values for start and stop.  The rationale behind eliminating the `start=None` keywords is that start always has a date specified within the test area, and lines after the `if start is None:` checks show up as missed lines in the code.

jklenzing:
Updates
- review comments
- tempfile and remove_files fully implemented in test_instruments, addressing #194.
- tempfile fully implemented in test_sw.
- documenting rationale in #194.
- found a few areas in the demos, documentation, and cosmic files where changes implemented in previous pulls somehow got dropped during the merge.
Still to do:
- remove_files not removing sw data from the tempfile directory.

jklenzing:
Also make sure that test_ftp_instruments is accounted for

jklenzing:
OK.  This is ready for review.  Addresses #194 and #198.  Added a few more tests to bump up coverage.

Given that the test routines delete data files, we should probably have a failsafe of some sort.  I'm fine holding off on merging until we can discuss at the next meeting.

pysat/ssnl/plot.py
rstoneback: Maybe?

Here is the code, it uses length but on a specific variable.

@property
    def empty(self):
        """Boolean flag reflecting lack of data.
        True if there is no Instrument data."""

        if self.pandas_format:
            return self.data.empty
        else:
            if 'time' in self.data.indexes:
                return len(self.data.indexes['time']) == 0
            else:
                return True

pysat/tests/test_instruments.py
rstoneback: I was thinking of something like that, perhaps 20 files or less? This check could lead to its own problems though.

 I agree with you that being in a temp directory that things should be fine. Can't help but be extra wary when deleting things in code that will run on somebody else's system.

pysat/_files.py
```diff
@@ -480,7 +480,7 @@ def from_os(cls, data_path=None, format_str=None,
         format_str : string with python format codes
             Provides the naming pattern of the instrument files and the
             locations of date information so an ordered list may be produced.
-            Supports 'year', 'month', 'day', 'hour', 'min', 'sec', 'version',
+            Supports 'year', 'month', 'day', 'hour', 'mins', 'sec', 'version',
```
rstoneback: This is very much responsive to my comment about min being a less than great variable choice. I'm not sure I feel the same way about its use as a format keyword. From a UI perspective 'mins' is now an odd keyword that breaks the naming pattern.

pysat/_meta.py
```diff
@@ -279,40 +279,6 @@ def keep(self, keep_names):
                 drop_names.append(name)
         self.drop(drop_names)
 
-#     def default_labels_and_values(self, name):
```
rstoneback: Thanks for clearing out some cruft

pysat/tests/test_files.py
```diff
@@ -121,29 +121,30 @@ def test_parse_delimited_filename(self):
         # Note: Can be removed if future instrument that uses delimited
         # filenames is added to routine travis end-to-end testing
         fname = ''.join(('test_{year:4d}_{month:2d}_{day:2d}_{hour:2d}',
-                         '_{min:2d}_{sec:2d}_v01_r02.cdf'))
+                         '_{mins:2d}_{sec:2d}_v01_r02.cdf'))
         year = np.ones(6)*2009
```
rstoneback: To keep some consistency, we could change year to years, month to months, for these variables. This is purely style now.

pysat/tests/test_files.py
```diff
@@ -121,29 +121,30 @@ def test_parse_delimited_filename(self):
...
-        dict = pysat._files.parse_delimited_filenames(file_list, fname, '_')
-        assert np.all(dict['year'] == year)
-        assert np.all(dict['month'] == month)
-        assert np.all(dict['day'] == day)
-        assert np.all(dict['hour'] == hour)
-        assert np.all(dict['min'] == min)
-        assert np.all(dict['day'] == day)
-        assert (dict['version'] is None)
-        assert (dict['revision'] is None)
+                                          mins=mins[i], sec=sec[i]))
```
rstoneback: yeah, the mins as a keyword seems weird to me. Open to discussion, of course, as a number of my original style choices have already been updated :)

pysat/tests/test_instruments.py
```diff
@@ -28,23 +28,15 @@
...
-    target = ''.join(('.', inst.files.files.values[0].split('.')[-1]))
-    for the_file in os.listdir(dir):
-        if (the_file.find(target) != -1):
-            file_path = os.path.join(dir, the_file)
-            if os.path.isfile(file_path):
-                os.unlink(file_path)
+    # remove any files downloaded as part of the unit tests
+    temp_dir = inst.files.data_path
+    for the_file in list(inst.files.files.values):
+        if the_file.rfind('_') > the_file.rfind('.'):
```
rstoneback: could you add some comments about what these lines are doing?

pysat/tests/test_meta.py
```diff
@@ -989,4 +963,15 @@ def test_transfer_attributes_to_instrument_strict_names(self):
         self.meta.transfer_attributes_to_instrument(self.testInst,
                                                     strict_names=True)
 
-        assert True
+    def test_merge_meta(self):
+        self.meta['new'] = {'units': 'hey', 'long_name': 'boo'}
+        meta2 = pysat.Meta()
+        meta2['NEW21'] = {'units': 'hey2', 'long_name': 'boo2',
```
rstoneback: :)

pysat/ssnl/plot.py
jklenzing: Looks like a straightforward replace.  May need to update the pandas_format=False case to look for
```
len(ds.variables)==0
```
which seems to be what xarray prefers.  This should be more generalized, ie, `time` does not have to exist.

jklenzing: created a new branch to see if this works.

pysat/tests/test_meta.py
jklenzing: The meta tests are the most fun to update

pysat/tests/test_instruments.py
jklenzing: Adding here and in test_sw.py

pysat/tests/test_files.py
jklenzing: There are arguments for both.  Adding to agenda.

pysat/ssnl/plot.py
jklenzing: The short answer is no.  :(
The data is not loaded to self until the end of the function, so the property cannot be used in Line 940 as the property does not yet exist.  But I think I found another way.

pysat/tests/test_files.py
jklenzing: An alternative is to just use 'minute', which would be less ambiguous.  Plural works fine for most of the keywords, but 'mins' and 'secs' aren't actually intuitive.

pysat/tests/test_files.py
jklenzing: Weird to me too.  :)

pysat/ssnl/plot.py
rstoneback: Time does have to exist, it is part of pysat's requirements on data. 

Replacing:
if len(inst.data) > 0:
with:
if not inst.empty:

doesn't work?

pysat/tests/test_files.py
rstoneback: Yeah, 'minute' and 'second' could work. All the names are now clear. I prefer short names but 'minute' is better than 'mins'.

rstoneback:
Hey, coverage is rounding up to 80%. We are back to a yellow testing badge! Thanks Jeff!

pysat/ssnl/plot.py
jklenzing: The `len(ds.variables) == 0` broke a bunch of other things.  I've left it as is.  

A simple replace doesn't work because the data object is a pure pandas or xarray object, not a pysat instrument, hence empty is not defined for xarray.  Check out Pull #199 for the fix.

jklenzing:
> Hey, coverage is rounding up to 80%. We are back to a yellow testing badge! Thanks Jeff!

Thanks!  That was my original goal of this pull request, but got sidetracked on a bunch of issues with making the testing easier.  Still have model_utils and sw_methods to go, but those are outside the scope here.

pysat/ssnl/plot.py
rstoneback: It is supposed to be a pysat Instrument. I do see that part is missing from the docstring.

jklenzing:
To do:
-[ ] Update 'mins' to 'minute', 'sec' to 'second'
-[ ] Add twenty-file check to delete routines

pysat/tests/test_files.py
jklenzing: Updated to minute and second.

pysat/tests/test_instruments.py
jklenzing: Added a 20 file check.  If 20 or more, no deletions occur and a warning message is issued.

jklenzing:
One additional change: scrubbed test_files.py to always use `''.join` for multi-line string concantenation.

jklenzing:
Update: The 20-file check works well for test_instruments, not so well for test_sw.  I think this is because the fake daily from monthly winds up with ~30 'files' per month, and f107.all having daily files back to 1947.  Not sure what to do here.  The advantage is that the sw data is very small, so leaving it in temp won't fill up the hard drive quickly.

pysat/_files.py
jklenzing: updated to minute and second

pysat/tests/test_instruments.py
aburrell: @jklenzing was the 'dir' variable name changed?

jklenzing:
@aburrell, I just searched the whole project and found that I missed one usage of `dir` as a variable name.  Updated and pushed.  Waiting for final test results from Travis CI.

jklenzing:
OK, final cleanup is due, pending more comments.  Ready for final review.

Issue #163, opened 2019-02-19
Default data directory path


aburrell:
The default data directory path puts everything in the user's home directory.  It might cause less problems for newby users to create a pysatData directory at the home directory level and use this as the default if nothing is specified.

jklenzing:
Should this be created during setup?  Or the first time data is downloaded?

rstoneback:
I believe the code that sets up the directory now is in the main init for pysat. It sets up a couple things the first time pysat is run.

On Mar 5, 2019, at 5:40 PM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


Should this be created during setup? Or the first time data is downloaded?

jklenzing:
I've got a version working on my local computer (branch is `default_directory`), but Travis apparently will not let me create a directory under the /home/travis directory, and the instrument downloads error out.  The quick solution is to put in a check for travis and send the data somewhere else.  Thoughts?

rstoneback:
You can create directories but those probably have to be within the user home.

The Travis.yml used to make some directories. May still have some examples.

On Mar 6, 2019, at 7:30 AM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


I've got a version working on my local computer (branch is default_directory), but Travis apparently will not let me create a directory under the /home/travis directory, and the instrument downloads error out. The quick solution is to put in a check for travis and send the data somewhere else. Thoughts?

jklenzing:
Closed with merge of #197

PR #197, opened 2019-03-07
Adds new default directory at home level

jklenzing:
Addresses #163.

Summary of changes
- Automatically assigns data to go into a 'pysatData' directory to be created at the user's home level.
- Prompt for user to set a new directory only occurs on initial load of pysat.
- Fixed a bug in test_utils.py that would not restore user's data directory after running unit tests locally.
- Removed `safe_data_dir` function in instrument tests, as this should not be needed now.

Notes
- I've opted for not actually creating the pysatData directory until after the first data download, largely because this is a pet peeve of mine with software packages.  This way, a potential user is given the chance to assign a directory before anything is created.  However, I'm willing to be persuaded otherwise.
- As part of the bug fix in test_utils, I've changed the assert statements for the two set_dir tests.  The goal of both tests is now to see if the initial change holds as well as whether the store option worked for a subsequent reload of pysat.


jklenzing:
Ready for review

PR #199, opened 2019-03-12
xarray len deprecation fix

jklenzing:
Addresses #130.

Starting with xarray 0.11.0, `len` cannot directly be run on a Dataset.  Updates syntax accordingly.

### DIFF
--- a/pysat/_custom.py
+++ b/pysat/_custom.py
@@ -126,7 +126,8 @@ class Custom(object):
         if len(self._functions) > 0:
             for func, arg, kwarg, kind in zip(self._functions, self._args,
                                               self._kwargs, self._kind):
-                if len(sat.data) > 0:
+                ind = sat.data.index if sat.pandas_format else sat.data.indexes
+                if len(ind) > 0:
                     if kind == 'add':
                         # apply custom functions that add data to the
                         # instrument object
--- a/pysat/_custom.py
+++ b/pysat/_custom.py
@@ -126,8 +126,7 @@ class Custom(object):
         if len(self._functions) > 0:
             for func, arg, kwarg, kind in zip(self._functions, self._args,
                                               self._kwargs, self._kind):
-                ind = sat.data.index if sat.pandas_format else sat.data.indexes
-                if len(ind) > 0:
+                if not sat.empty:
                     if kind == 'add':
                         # apply custom functions that add data to the
                         # instrument object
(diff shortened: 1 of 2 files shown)


pysat/_custom.py
```diff
@@ -126,7 +126,8 @@ def _apply_all(self, sat):
         if len(self._functions) > 0:
             for func, arg, kwarg, kind in zip(self._functions, self._args,
                                               self._kwargs, self._kind):
-                if len(sat.data) > 0:
+                ind = sat.data.index if sat.pandas_format else sat.data.indexes
```
rstoneback: I think this should be if not sat.empty:

pysat/_instrument.py
```diff
@@ -985,7 +985,8 @@ def _load_data(self, date=None, fid=None):
             raise TypeError('Metadata returned must be a pysat.Meta object')
 
         # let user know if data was returned or not
-        if len(data) > 0:
+        ind = data.index if self.pandas_format else data.indexes
+        if len(ind) > 0:
```
rstoneback: same.

I really want to keep checks on data type to the absolute minimum. Folks, including us, shouldn't have to know ahead of time what kind of data is coming down.

pysat/_custom.py
jklenzing: I agree, but we need to create a routine that adds an `empty` property to the xr.Dataset.

rstoneback: I don't think so, that is the point of the 'empty' attribute attached to each pysat.Instrument object. It internally deals with the differences between pandas and xarray.

jklenzing: You're right.  It's the `_instrument` line I'm thinking of.

pysat/_instrument.py
jklenzing: This is the bit that has a data object that becomes attached to the pysat.instrument.  The data object (if xarray) does not have an empty property and won't until after this routine is run.

pysat/_custom.py
jklenzing: fixed here

pysat/_instrument.py
```diff
@@ -922,7 +922,7 @@ def _load_data(self, date=None, fid=None):
 
         Returns
         --------
-        data : (pds.DataFrame)
+        data : (pds.DataFrame or xr.Dataset)
```
rstoneback: Thanks for catching this.

Issue #187, opened 2019-03-03
Consider version restrictions


rstoneback:
Python's setup.py supports version ranges. Given that packages continue to evolve, we may want to pair pysat to library versions that we've tested with. Some packages, particularly pandas, can implement changes with new versions that can break pysat.

jklenzing:
Looking over the latest version of pandas (0.24.0), it looks like this may be a good candidate for an upper limit.  The only deprecation I recognize in this list is the Series.to_csv one, which we updated in #173.  Looks like Panel is still active, but probably should run some checks.  This version is noted to be the last that will be backwards compatible with Python 2.7,

https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.24.0.html

jklenzing:
For the record, I've been running pandas 0.23.4 while developing my unit tests, so we know this is safe.

jklenzing:
Ran unit tests for `ssnl.avg` and 'cosmic_and_ivm_demo.py' successfully under pandas 0.24.1.

jklenzing:
From http://xarray.pydata.org/en/stable/whats-new.html

xarray v0.11.0 will be the last major release to support python 2.7.
xarray v0.12.0 will deprecate the `inplace` keyword from 'number of DataArray and Dataset methods'

rstoneback:
Looks like v.11.0 will be our version for a bit. Pysat needs to maintain great compatibility with python 2. The ICON and COSMIC-2 projects have pysat running in some data centers for some time.

On Mar 11, 2019, at 10:25 PM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


From http://xarray.pydata.org/en/stable/whats-new.html

xarray v0.11.0 will be the last major release to support python 2.7.
xarray v0.12.0 will deprecate the inplace keyword from 'number of DataArray and Dataset methods'

rstoneback:
Plan is to put in a cap at current versions to help mitigate the chances that something breaks with new releases. We will have to monitor it. Perhaps remove it when things are more stable.

rstoneback:
Apply version restrictions at the master branch level so we can see problems.

jklenzing:
Speaking of which, Travis CI just upgraded to xarray 0.12.0, and the python 2.7 tests fail pretty much immediately.

jklenzing:
[...]

PR #201, opened 2019-03-16
version limits

jklenzing:
Addresses #187. 

Adds version limits to pandas and xarray for python 2.7. Python 3 and up has no limit largely so Travis CI can keep testing the latest version and update us on when new versions break.


aburrell:
Ah, wait!  Should there be an upper limit on pandas for both 2 and 3 so that panel still works?

jklenzing:
> Ah, wait! Should there be an upper limit on pandas for both 2 and 3 so that panel still works?

Pandas still has not announced which version panel will be removed from.

aburrell:
Ok, I stand by my approval.  Go ahead and merge.

Issue #202, opened 2019-03-17
Reduce code duplication in testing instruments


jklenzing:
There are several areas in the testing instruments that can be combined to reduce duplication of code.  
- `testaddx` instruments can be folded into the main pysat_testing.py instrument, using tags to change data output.
- `testsmall` instruments can be potentially be folded back into pysat_testing.py, as `tag='100'` would create default behaviour.
- With current parameters, the test constellations `test_diff`, `test_diff2`, and `test_diff_small` should perform the exact same function.  @rstoneback, is this a placeholder for future tests, or can these be combined?

jklenzing:
Closing with merge of #203.

Issue #196, opened 2019-03-06
test_sw should utilize test data


jklenzing:
Current behaviour of test_sw includes required data download for some tests.  These tests are mostly skipped on Travis CI due to FTP usage of some of the datasets.

@aburrell, is download required for these methods?  Data that is available via HTTPS should be already tested for in the test_instruments routine.   If we put together some fake data, this will improve Travis coverage and minimize the tests we need to run locally in addition to the remote tests.  Or is there something I'm missing?

jklenzing:
Path forward: Adding downloaded data to test_data directory.  About 440 KB total, so minimal to project.

jklenzing:
Closing with merge of #203

Issue #46, opened 2018-07-07
pysat testing improvement


rstoneback:
The simple satellite simulations (instruments/pysat_testing.py) repeatedly use a np.mod function to create simple periodic signals (local time, longitude, etc.). These calls should be replaced with a function in the same file.

aburrell:
Some of this may have been addressed in the 'models' branch.

jklenzing:
Closing with merge of #203.

PR #203, opened 2019-03-19
Unit Testing Infrastructure Upgrade

jklenzing:
Addresses #46, #196, and #202.

Summary of Changes:
- Restructured `test_sw` to use test data rather than download (#196).
  - Added relevant kp and f107 data to test_data directory
  - Unit tests use a fixed test_day rather than today for testing
  - Added test_data_path to pysat.__init__
  - Removed skip_tests, tempfile notation.
- Reduced code duplication by adding functionality to pysat testing instruments (#202).
  - Added tags to pysat_testing to change the default behaviour of `dummy1`.
  - Old `tag` behaviour (total number of data points) for numeric strings migrated to `sat_id`
  - Various tests invoking numeric string as `tag` updated.
  - testaddx and testsmall instruments removed (replaced with pysat_testing)
  - `_generate_fake_data` function added to remaining 4 testing instruments (#46).
  - Updated test_constellation objects to use tag structure to determine `dummy1` variable
- General bug and style fixes
  - Fixed a bug in `_meta.keep`
  - Updated style in test_omni_data to make more readable
  - Updated some tests in test_ssnl_occur_prob to be more resilient

This is the second half of the testing infrastructure upgrade I started in #192.  Travis CI tests are down to 40 min.


jklenzing:
Open Issue:  The `calculate_imf_steadiness` routine in omni_hro.py looks like the test object should be an omni_hro instrument, but unit tests currently use a pysat_testing object.  Not sure if I am interpreting the function correctly.

jklenzing:
Another open issue: `_generate_fake_data` could be placed in a common area like `utils` rather than in every test file.

jklenzing:
New updates to clean up code:
- `generate_fake_data` has been moved to `testing_methods.py` to further reduce duplication.  This will be moved to `methods.testing` as part of #214.

Note that ~26,462 of the 26,769 added lines are the test dataset added to speed up the test time.

pysat/_meta.py
```diff
@@ -272,7 +272,7 @@ def drop(self, names):
     def keep(self, keep_names):
         """Keeps variables (keep_names) while dropping other parameters"""
 
-        current_names = self._data.columns
+        current_names = self._data.index
```
aburrell: What instigated this change?

pysat/instruments/pysat_testing2d.py
```diff
@@ -86,27 +87,25 @@ def load(fnames, tag=None, sat_id=None):
...
     # create a fake longitude, resets every 6240 seconds
     # sat moves at 360/5820 deg/s, Earth rotates at 360/86400, takes extra time
     # to go around full longitude
-    long_uts_root = np.mod(time_delta.total_seconds(), 6240)
-    longitude = np.mod(long_uts_root + num_array, 6240) * (360. / 6240.)
-    data['longitude'] = longitude
-
+    data['longitude'] = test.generate_fake_data(time_delta.total_seconds(),
+                                                num_array, period=6240,
+                                                data_range=360.0)
```
aburrell: having data range be a single number seems like a bad idea.  Shouldn't it have a max and min?  E.g., data_range=[0.0, 360.0]

pysat/instruments/testing_methods.py
```diff
@@ -0,0 +1,31 @@
...
+        For cyclic functions, the range of data values cycled over one period.
+        Not used for non-cyclic functions.
+        (default = 24.0)
+    cyclic : bool
+    """
+
+    if cyclic:
+        uts_root = np.mod(t0, period)
+        data = (np.mod(uts_root + num_array, period)
+                * (data_range / float(period)))
```
aburrell: There seems to be an implicit assumption that we will always want to start with zero, and I can envision cases where that will not be true.

pysat/tests/test_constellation.py
```diff
@@ -248,8 +241,7 @@ def test_diff_similar_instruments(self):
                                         cost_function)
         diff = results['dummy1']
         dist = results['dist']
-        for i in diff:
-            assert i == 5
+        assert abs(diff - 5).max() == 0
```
aburrell: This is a different test than the one it replaced.  Is there a reason?  Why not use np.all([i == 5 for i in diff])?

pysat/tests/test_constellation.py
```diff
@@ -264,7 +256,8 @@ class TestDataMod:
     def setup(self):
         '''Runs before every method to create a clean testing setup.'''
```
aburrell: Need to change the single quotation marks to double quotation marks.

pysat/_meta.py
jklenzing: `keep` and `drop` should be similar functions.  I believe that both should act on the index in order to keep or drop metadata names (`keep` actually calls `drop`, so an inconsistency results in a KeyError).

test code:
```
import pysat
test = pysat.Instrument('pysat', 'testing')     
test.load(date=pysat.datetime(2009,1,1))  
test.meta.data  
```
yields
```
                                     units            long_name  ... custom Bin_Location
uts                                      s       Universal Time  ...  False          NaN
Epoch          Milliseconds since 1970-1-1                Epoch  ...    NaN          0.5
mlt                                  hours  Magnetic Local Time  ...    NaN          NaN
slt                                  hours     Solar Local Time  ...    NaN          NaN
orbit_num                                          Orbit Number  ...    NaN          NaN
longitude                          degrees            Longitude  ...    NaN          NaN
latitude                           degrees             Latitude  ...    NaN          NaN
dummy1                                                   dummy1  ...    NaN          NaN
dummy2                                                   dummy2  ...    NaN          NaN
dummy3                                                   dummy3  ...    NaN          NaN
dummy4                                                   dummy4  ...    NaN          NaN
string_dummy                                       string_dummy  ...    NaN          NaN
unicode_dummy                                     unicode_dummy  ...    NaN          NaN
int8_dummy                                           int8_dummy  ...    NaN          NaN
int16_dummy                                         int16_dummy  ...    NaN          NaN
int32_dummy                                         int32_dummy  ...    NaN          NaN
int64_dummy                                         int64_dummy  ...    NaN          NaN
```
Calling 
```
test.meta.drop('int64_dummy')    
```
yields
```
                                     units            long_name  ... custom Bin_Location
uts                                      s       Universal Time  ...  False          NaN
Epoch          Milliseconds since 1970-1-1                Epoch  ...    NaN          0.5
mlt                                  hours  Magnetic Local Time  ...    NaN          NaN
slt                                  hours     Solar Local Time  ...    NaN          NaN
orbit_num                                          Orbit Number  ...    NaN          NaN
longitude                          degrees            Longitude  ...    NaN          NaN
latitude                           degrees             Latitude  ...    NaN          NaN
dummy1                                                   dummy1  ...    NaN          NaN
dummy2                                                   dummy2  ...    NaN          NaN
dummy3                                                   dummy3  ...    NaN          NaN
dummy4                                                   dummy4  ...    NaN          NaN
string_dummy                                       string_dummy  ...    NaN          NaN
unicode_dummy                                     unicode_dummy  ...    NaN          NaN
int8_dummy                                           int8_dummy  ...    NaN          NaN
int16_dummy                                         int16_dummy  ...    NaN          NaN
int32_dummy                                         int32_dummy  ...    NaN          NaN
```

Calling 
```
test.meta.keep('uts')
```
should yield
```
                                     units            long_name  ... custom Bin_Location
uts                                      s       Universal Time  ...  False          NaN
```
In the current develop branch, this results in a KeyError, since 'uts' does not appear in the list of columns.

I could be misreading the code, and drop should be the one to be changed so that both operate on the columsn.  @rstoneback?

pysat/instruments/pysat_testing2d.py
jklenzing: Adding

pysat/instruments/testing_methods.py
jklenzing: Will add.

pysat/tests/test_constellation.py
jklenzing: fixed here and elsewhere

pysat/tests/test_constellation.py
jklenzing: Should be functionally equivalenet, but will replace with assert np.all(abs(diff - 5)) == 0

Also deleted dist = results, since this is not used in the test

pysat/_meta.py
rstoneback: You have it correct, the keep should drop of all of the other variable names, stored in the index.

rstoneback:
Thanks for all of the code!

PR #212, opened 2019-04-22
Update sw_f107.py

aburrell:
Bug for new downloads, because there is never enough testing.


aburrell:
Feel free to review Jeff, but I will test this once more locally tomorrow to be sure it works as intended with incremental updates to existing files.

jklenzing:
Don’t see anything out of place here. Will wait for local tests as this is hidden from the Eye of Travis.

aburrell:
And NOAA is still down.

aburrell:
Ok @jklenzing, local success on day two.  Give it a go.

PR #220, opened 2019-05-20
bugfix - saber directory

jklenzing:
CDAWeb has updated the directory structure for TIMED/SABER to add new data products.  Rather than add everything now, I'm updating the default dataset so that unit tests can continue uninterrupted.


jklenzing:
New directory path for reference: https://cdaweb.gsfc.nasa.gov/pub/data/timed/saber/level2a_v2_00_CDF/

Issue #219, opened 2019-05-14
pandas sorting behavior in append


jklenzing:
The pandas behavior for sorting appended data changed default behavior in a future version of pandas.  This required adding the `sort=True` keyword to append and other concat routines, but this keyword was added in pandas version 0.23, and therefore breaks earlier versions.

https://pandas.pydata.org/pandas-docs/version/0.23/generated/pandas.DataFrame.append.html

Propose updating minimum pandas level to 0.23, which does support python 2.7.

Previous changes and rationale documented in #125 and #152.

jklenzing:
Closed with merge of #221.

PR #221, opened 2019-05-22
bugfix for #219

jklenzing:
Added minimum support for pandas to address #219.

Also updated setup info to include python 3.7.


rstoneback:
I just found out through error that we need pandas 0.23. That's what I get for getting behind on pulls. Thanks again @jklenzing.

Issue #222, opened 2019-05-26
supermag tests not downloading in unit tests.


jklenzing:
The supermag tests in test_instruments.py are failing in Travis and locally.  If I try to manually download the test data in ipython, there is no problem.  Likely not a server issue, but a unit test issue.

```
======================================================================
FAIL: Checking load routine functionality for module with clean level "none":  supermag magnetometer indices 
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/rstoneback/pysat/pysat/tests/test_instruments.py", line 290, in check_load
    assert not inst.empty
AssertionError

======================================================================
FAIL: Checking load routine functionality for module with clean level "none":  supermag magnetometer  
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/rstoneback/pysat/pysat/tests/test_instruments.py", line 290, in check_load
    assert not inst.empty
AssertionError

======================================================================
FAIL: Checking load routine functionality for module with clean level "none":  supermag magnetometer all 
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/rstoneback/pysat/pysat/tests/test_instruments.py", line 290, in check_load
    assert not inst.empty
AssertionError
```

jklenzing:
See logs in https://travis-ci.org/rstoneback/pysat/builds/537227030

jklenzing:
Confirmed fails limited to clean_level="none"

jklenzing:
requests is retrieving the text "securityerror" rather than a data file.  Not familiar enough with supermag to fix this.  @aburrell? @rstoneback?

jklenzing:
> Confirmed fails limited to clean_level="none"

Followup: downloads work, but not loads.

rstoneback:
[...]

PR #223, opened 2019-05-27
Supermag download

jklenzing:
Addresses #222.

Supermag no longer supports download via script.  Removing `download` routine and replacing with warning message to direct users to download site, similar to champ_star.

Also updated docstring and supported keywords in `list_remote_files` for consistency.


rstoneback:
I don't want to delete the code. I do like the warning though. Perhaps we should rename the original download method in hopes that SuperMAG will do better in the future.

jklenzing:
We can always pull the code back from an earlier commit.  I don't like removing this, but it comes down to ~200 lines of unused, untested code.  Since we can't do end-to-end testing on supermag, we've already lost most of the coverage on this file (80% down to 13%), and total coverage is back below 80%.

We could also comment it out, which in general we have been trying to avoid in previous pull requests.  @aburrell?

jklenzing:
There's a version up now where the old download routine is maintained as `test_download`, in case we want to test the downloads in the future.  Code coverage drops 0.8% to 77.47%.

rstoneback:
So it goes with testing coverage.

I think our download warning text should be stronger. SuperMAG made a choice to block downloads.

“SuperMAG and the Johns Hopkins Applied Physics Laboratory have chosen to block downloads using any tool other than the website (http://supermag.jhuapl.edu) and direct human interaction. “

Could add “ Please contact the National Science Foundation (find sponsor email) if you disagree with restricted access for public data.” Once I find the correct program /officer email.

On May 28, 2019, at 2:17 AM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


There's a version up now where the old download routine is maintained as test_download, in case we want to test the downloads in the future. Code coverage drops 0.8% to 77.47%.

jklenzing:
I think that until we have more info, the current warning is sufficient.  We should get this merged to fix the unit tests for now.

rstoneback:
That’s fine.


On May 28, 2019, at 4:29 PM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


I think that until we have more info, the current warning is sufficient. We should get this merged to fix the unit tests for now.

Issue #233, opened 2019-06-04
SuperMAG download options


jklenzing:
Documenting our changes for the old SuperMAG download routine (called `test_download` as of the #223 merge).  History of issue documented in #222 and #223.

The main issue right now is that this is 100+ lines of untested and unusable code since SuperMAG blocked script access.  

Potential courses of action:
- delete function and point to version 1.2 to resurrect download routines in case script access is restored.
- comment out code so that version is not lost in the git labyrinth.
- leave things as they are and take the ~0.8% hit in the testing.

jklenzing:
My preferred course is to delete and point to the last usable version.  @aburrell? @rstoneback?

rstoneback:
My preference is to comment this time since it reduces the chances we forget in the future. That being said I can absolutely accept deleting it and relying on git to get the code back if SuperMAG opens data back up.

On Jun 4, 2019, at 10:36 AM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


My preferred course is to delete and point to the last usable version. @aburrell<https://github.com/aburrell>? @rstoneback<https://github.com/rstoneback>?

jklenzing:
Angeline votes for delete and point to old version.

jklenzing:
Note for future us:  The last copy of the SuperMAG download routine is the `test_download` function in commit 1f21f8afd06c558fc8bc08c500e3c377c4a879bc

See you on the other side, old code.

jklenzing:
Closed with merge of #235.

PR #235, opened 2019-06-05
Supermag Cleanup

jklenzing:
Addresses #233.  

Removes the old download routine (test_download) since it currently cannot be used with SuperMAG server and is therefore untested.  Pointing to last commit with this function in the issue.


jklenzing:
Between this and #234, the coverage badge should be back to yellow.

jklenzing:
It lives on in the git.

PR #236, opened 2019-06-05
Unit tests bugfix for sw_kp

jklenzing:
Caught a major bug in the unit tests that dropped some data products from testing, so pushing this sooner rather than later.

Summary of Changes:
- Found a bug where the unit tests skipped downloads for sw_kp.  The `break` statement at line 100 in test_instruments.py kicks out of the loop over tags, so if one tag is excluded for an instrument, all tags would be excluded.  Restructured as an if-else statement so these tests are not skipped.
- Added all tags to `sat_ids` in sw_kp.
- Added two unit tests for meta.
- Cleaned up some comments.


pysat/tests/test_meta.py
```diff
@@ -20,6 +20,10 @@ def teardown(self):
         del self.testInst
         del self.meta
 
+    @raises(ValueError)
+    def test_setting_nonpandas_metadata(self):
+        self.meta = pysat.Meta(metadata='Not a Panda')
```
rstoneback: but I like pandas! Do you think if it tries hard enough it can become a panda? Like pinocchio?

jklenzing: 🐼

PR #231, opened 2019-06-03
Orbit iteration fix when no loaded data (non-mlt orbits)

rstoneback:

PR #239, opened 2019-06-14
Update timed_saber.py

jklenzing:
Fixes a bug found in #231. 

CDAweb renamed the core TIMED_SABER directory for the 2.00 data.


jklenzing:
Minor update to path, passing unit tests.  Merging based on self-review.

PR #248, opened 2019-07-05
update to pull request template

jklenzing:


jklenzing:
Travis tests are failing due to pandas update (see #249).  Since this is a documentation update only, it should be fine.

PR #249, opened 2019-07-05
Limit pandas to < 0.25

jklenzing:
The pandas apocalypse is at hand!

Pandas 0.25 has removed panel.  Adding an upper limit to pandas version to keep the code moving.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ if sys.version_info.major == 2:
                         'pysatMagVect', 'madrigalWeb', 'h5py',
                         'PyForecastTools', 'pyglow']
 else:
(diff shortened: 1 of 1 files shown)


rstoneback:
The seasonal code is the part that may actually encounter a panel. Since we are breaking those parts out, this problem will get sorted out as a by product.

jklenzing:
The actual error I'm getting (in the ssnl tests) is 
```
AttributeError: type object 'Panel' has no attribute 'from_dict'
```
 According to https://pandas-docs.github.io/pandas-docs-travis/whatsnew/v0.25.0.html
`Panel` is / will be removed.  I think that Travis is pulling a dev version of 0.25.0.

Issue #262, opened 2019-08-15
multi_file_day option reveals concat_data bug


rstoneback:
**Describe the bug**
sort=True is a keyword on line 704 of _instrument.py, within concat_data. Needs to be removed and sorting behavior investigated now that option has been deprecated.

**To Reproduce**
1. create an Instrument object with multi_file_day=True
2. Load a day of data with neighboring days of data
3. See error

**Expected behavior**
We(I) should've had a test with this flag to catch any pandas version deprecations.

**Screenshots**

**Desktop (please complete the following information):**
 - Other details about your setup that could be relevant

**Additional context**

rstoneback:
See pull #264

jklenzing:
`sort` is a new option with pandas 0.23.  The "deprecation" comes from using an old version of pandas.  It looks like the issue is coming from using pandas 0.20 in pull #264.  I think this means we need to update the pandas version rather than remove the option.

See #125 and #152 for rationale for adding this option.  Essentially it was added because the default sorting behaviour was changing.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html

rstoneback:
Thanks Jeff. My old pandas version got me in multiple places. I’ll update it back.


On Aug 21, 2019, at 4:46 PM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:


sort is a new option with pandas 0.23. The "deprecation" comes from using an old version of pandas. It looks like the issue is coming from using pandas 0.20 in pull #264<https://nam02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Frstoneback%2Fpysat%2Fpull%2F264&data=02%7C01%7Crstoneba%40utdallas.edu%7C9cec8661fb4a49f5c93e08d72680fb47%7C8d281d1d9c4d4bf7b16e032d15de9f6c%7C0%7C0%7C637020207730112009&sdata=%2FFk%2BR8gK3yhfcn6vweg5gkJA5l0fTZcIjTxyah9aWgs%3D&reserved=0>. I think this means we need to update the pandas version rather than remove the option.

[...]

PR #261, opened 2019-08-14
Variable name case standards 

rstoneback:
# Description
pysat began life with an assumption that all variable names would be lower case. Earlier versions imposed this, I believe. Work on ICON has forced some changes to the code to support names with non-lowercase letters, seen most explicitly in the Meta object. Within meta, the first instance of a string sets the case, any later variations are treated as the original variable. This is particularly good when one person thinks an attribute should be 'units', and another 'Units', or another 'UNITS'. Meta treats them all the same. This significantly helps Meta maintain compatibility with a larger number of file and data standards. Particularly in the context of community development.

For ICON, many variables have mixed case.  The IVM software previously made everything lower case while in-house, then made everything upper case for the output file. New standards for level-2 files have long filenames, underscores, and upper and lower letters. 

What should pysat standards be?
1) Accept variable names as is. Has the problem that Meta is case-insensitive. So the data could have two variables, 'mlt', and 'MLT', with only one Metadata entry.
2) Force all data variables to upper case. No.
3) Force all data variable names to lower case. Fancier name stored in Metadata so output file names are supported. We would need to make some changes within pysat to support this.
   - Update to_netCDF to use metadata names. Write new tests around this.
   - Update from_netCDF to use lower case names on the data side, preserve case on Meta side.
   - Modify the setitem method to ensure variable names are always lower case
   - Create a Instrument.data attribute that checks variables names any time someone sets their own data set within an Instrument object, e.g. inst.data = new_data , to make sure all variable names are lower case.

Right now we are pretty much 1, though there may be some residual code floating around. 

This pull request is intended to start a conversation. (See update below - I now propose a more limited pull)

Addresses # 260

Adds new file attributes when using to_netCDF4 so that compatibility with ICON file standards is maintained.

Makes first step to supporting option 3 above. Uses the variable names within the Meta object when naming variables within the netCDF file.
 - Update: Functionality has been implemented as an option when writing a file. I propose remaining at option 1 as it involves the least user confusion. Variable names are taken as given by a user or data set. If a user decides, they can now exploit more of pysat's functionality with the file write option here.

## Type of change

Please delete options that are not relevant.

- New Feature (updated since original pull)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for
your test configuration

- Created files using the ICON IVM software. 
[...]

PR #264, opened 2019-08-21
Concat data update - addresses issue #262

rstoneback:
# Description

Addresses # 262

Removes deprecated argument from pandas' concat method. 
Added a test to ensure method is regularly run.

## Type of change

# How Has This Been Tested?

I encountered the bug using the multi_file_day option while processing COSMIC-2 data. Pandas informed me that sort=True no longer supported. 

**Test Configuration**:
* Operating system
* Version number pandas 0.20.
 I was below setup.py reqs. Apparently python setup.py develop doesn't really enforce requirements. Issue that cropped up due to a development environment with heritage.

# Checklist:

- [X] Make sure you are merging into the ``develop`` (not ``master``) branch
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation (No changes needed)
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules
- [X] Add a note to ``CHANGELOG.md``, summarizing the changes


pysat/_instrument.py
```diff
@@ -701,7 +701,7 @@ def concat_data(self, data, *args, **kwargs):
         """Concats data1 and data2 for xarray or pandas as needed"""
 
         if self.pandas_format:
-            return pds.concat(data, sort=True, *args, **kwargs)
+            return pds.concat(data, *args, **kwargs)
```
rstoneback: This is the relevant change.

pysat/tests/test_instrument.py
```diff
@@ -174,6 +174,30 @@ def test_filtered_date_attribute(self):
         self.testInst.date = now
         assert today == self.testInst.date
 
+
```
rstoneback: Relevant changes below.

rstoneback:
This pull incorrectly includes code from pull #261

CHANGELOG.md
```diff
@@ -2,14 +2,18 @@
...
 
-## [2.X.X] - 2019-07-17
+## [2.X.X] - 2019-08-20
 - New Features
    - Added new velocity format options to utils.coords.scale_units
    - Improved failure messages for utils.coords.scale_units
    - Added some tests for model_utils
+   - Added option to to_netCDF that names variables in the written file
+   based upon the strings in the Instrument.meta object
+   - Improved compatibbility with NASA ICON's file standards
```
rstoneback: Typo fixed in other pull.

jklenzing:
See comments in #262 before merging.  `sort` is a new feature, rather than a deprecated one.

rstoneback:
I fixed the code back up, restoring sort=True. Now this pull only adds a test for the concat_data method.

rstoneback:
Thanks for reminding me of the code history.

After a bit more time with the code I think we should go back to leaving sort=True out. The concat_data Instrument method is otherwise set up to pass extra args and keyword args through to the underlying concat methods. Fixing one argument seems inconsistent.

I do appreciate that sort=True was put in to maintain old behaviors. That being said, the places in pysat where it is used doesn't rely on sort=True. The test I wrote, as part of this pull, no longer works since the columns are in a different order than the original data. I could change up the tests, but now we have different behaviors for pandas and xarray.

jklenzing:
I'm not opposed to pulling this out.  We just need to make sure that the code does what we expect.  As of pandas 0.25.0: "The current default of sorting is deprecated and will change to not-sorting in a future version of pandas."

An advantage to leaving it out is that xarray does not (currently) have this as a keyword.

http://xarray.pydata.org/en/stable/generated/xarray.concat.html

rstoneback:
I was thinking along the same lines. It's good to have some consistency across our data formats.

I will poke at the behavior in a few more situations and report back.

pysat/tests/test_instrument.py
rstoneback: This comment no longer needed now that the ICON updates have been properly merged. All code differences are authentic now.

rstoneback:
This is ready for review. Added @jklenzing as he already chimed in (Thanks!) due to recent work in this area. The default behavior is the same across datasets, pandas and xarray. For pandas, this requires setting sort to False, and for xarray specifying the time dimension. Support was also added so users can pass in their own values for these keyword arguments.

Since this pull was long, I can summarize. I ran into an issue caused by an out of date version of pandas. The pull was then redirected to improve consistency and add unit tests to the concat_data_method instead. A data access issue was encountered with xarray in the unit tests and was fixed by @asher-pembroke.

CHANGELOG.md
```diff
@@ -19,7 +19,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Bug fix
    - Fixed implementation of utils routines in model_utils and jro_isr
    - Fixed error catching bug in model_utils
+   - Removed deprecated keyword from methods within Instrument.concat_data
```
jklenzing: doesn't describe latest changes

pysat/tests/test_instrument.py
```diff
@@ -174,6 +174,59 @@ def test_filtered_date_attribute(self):
...
+        # data set #1
+        data1 = self.testInst.data
+        len1 = len(self.testInst.index)
+        # concat together
+        self.testInst.data = self.testInst.concat_data([data1, data2])
+        # test for concatenation
+        len3 = len(self.testInst.index)
+        assert (len3 == len1 + len2)
+
+        # concat together with sort=True
```
jklenzing: is this comment for both `if` and `else`?

Issue #115, opened 2018-11-09
C/NOFS IVM Cleaning Routine


jklenzing:
C/NOFS IVM cleaning routine uses one level of cleanliness for all data, but reliability criteria for drifts is more stringent than density values.  We should update accordingly.

jklenzing:
Proposed cleaning mechanism:
- Use a baseline for density measurements (based on rpaflag), replace more stringent parameters (such as drifts) with NaNs.
- Ion temperatures should require a O^+ fraction of 0.15 or higher, citing the Hairston paper.
- From my notes, drifts should be based on a combination of [O^+] and flag information.  In my old matlab routines, there is a 'dmpts' parameter in addition to the drift meter flag which I don't see in the pysat metadata.  It's likely my shorthand for something rather than an official name.
- I would also suggest dropping the drifts for 100% O^+.

rstoneback:
We will keep it simple, different parameters for cleaning on drifts, lower threshold for density, temperature, and composition.

aburrell:
I think it would be useful to look at the DEMETER studies for guidance on identifying useful conditions for CINDI drifts.

jklenzing:
Updates:  Reading over the documentation, it looks like the above parameters are handled in the quality flags.  However, to close this issue, there should be a way to be more stringent on velocity (rpa flag of 0 only) whilst letting the density and other parameters use more flags (rpa flag 0 or 1).  

The simplest way to implement this is to just pass NaNs through for velocity values where the rpa flag is 1 on the 'clean' setting.  Does anyone see any issues with this?

aburrell:
That's fine with me.  If I recall correction, I think that in 'dusty' velocities should be passed down to RPA flag of '3'.  Aren't the other parameters also good down to the RPA_flag=3?  Or am I remembering the IDM flag?

jklenzing:
From the document on CDAWeb:

RPA Quality Flag
0.  High quality data. 
1.	High quality data except for the ram velocity Vx and the spacecraft sensor plane potential. These two values are suspect and should be used with caution.
2.	Not used.
3.	Good quality data. The RMS fitting error is large and/or |Vx|>1000 m/s. The data may have more noise than quality flag = 0 data. Use Vx with caution.
4.	Good quality data. The RMS fitting error may be large, leading to greater noise levels. The ram velocity Vx and the spacecraft sensor plane potential values are suspect and should not be used.
5.	Undetermined. This value will only be seen if the quality control algorithm has failed.
[...]

PR #289, opened 2019-09-20
ENH: C/NOFS IVM Cleaning Routine Update

jklenzing:
# Description

Addresses #115.

Currently the C/NOFS IVM cleaning routine only provides data for where the drifts are reliable.  However, it is possible in some cases to have reliable density, composition, and temperature information where the drifts are suspect.  This updates the clean parameters to allow all desired data through while replacing suspect drifts with NaNs.

Instrument quality flags described in https://cdaweb.gsfc.nasa.gov/pub/data/cnofs/cindi/CINDI%20IVM%20Quality%20Flag%20Description.doc (last updated 2018-01-25).

## Type of change

# How Has This Been Tested?

tested ivm.load locally.

# Checklist:

### DIFF
--- a/pysat/instruments/cnofs_ivm.py
+++ b/pysat/instruments/cnofs_ivm.py
@@ -121,14 +121,15 @@ def clean(inst):
     idx, = np.where(inst.data.RPAflag <= max_rpa_flag)
     inst.data = inst[idx, :]
 
-    # Second pass, find bad drifts
+    # Second pass, find bad drifts, replace with NaNs
     idx, = np.where(inst.data.driftMeterflag > max_dm_flag)
 
-    # Also exclude very large drifts
+    # Also exclude very large drifts and drifts where 100% O+
     if (inst.clean_level == 'clean') | (inst.clean_level == 'dusty'):
         try:
             idx2, = np.where(np.abs(inst.data.ionVelmeridional) >= 10000.)
-            idx = np.unique(np.append(idx, idx2))
+            idx3, = np.where(inst.data.ion1fraction >= 1.0)
+            idx = np.unique(np.append(idx, idx2, idx3))
         except AttributeError:
             pass
 
@@ -136,6 +137,11 @@ def clean(inst):
                     'ionVelocityX', 'ionVelocityY', 'ionVelocityZ']
     for label in drift_labels:
         inst[label][idx] = np.NaN
+        
+    # Check for bad temperature fits (O+ < 15%), replace with NaNs
+    # Criteria from Hairston et al, 2015
+    idx, = np.where(ivm.data.ion1fraction < 0.15)
+    inst['ionTemperature'][idx] = np.NaN
 
     # basic quality check on drifts and don't let UTS go above 86400.
     idx, = np.where(inst.data.time <= 86400.)
--- a/pysat/instruments/cnofs_ivm.py
+++ b/pysat/instruments/cnofs_ivm.py
@@ -121,8 +121,8 @@ def clean(inst):
     idx, = np.where(inst.data.RPAflag <= max_rpa_flag)
     if (inst.clean_level == 'clean'):
         nO = inst.data.ion1fraction*inst.data.Ni
-        idx2 = np.where((inst.data.RPAflag <= 3) and (nO > 3.0e4))
-        idx = np.unique(idx.extend(idx2))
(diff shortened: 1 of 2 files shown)


jklenzing:
Hmm.  Slight decrease in coverage due to not covering final `else` statement.

https://coveralls.io/builds/25831087/source?filename=pysat%2Finstruments%2Fcnofs_ivm.py#L127

Shouldn't testing clean_level=None wind up here?

pysat/instruments/cnofs_ivm.py
```diff
@@ -113,31 +113,39 @@ def clean(inst):
     # make sure all -999999 values are NaN
```
aburrell: The altitude limit for clean should be removed

rstoneback:
If clean level is set to None the clean method shouldn’t be called. That’s what I remember at least. On phone.

jklenzing:
> If clean level is set to None the clean method shouldn’t be called. That’s what I remember at least. On phone.

Makes sense.

jklenzing:
Fixed the 'dirty' vs None case, removed altitude limits.  Will produce some examples of data for the record.  I'm open to adding the density and composition limits, but we should work out the exact limits first.

From my Mar 5 notes in the issue:
- Ion temperatures should require a O^+ fraction of 0.15 or higher, citing the Hairston paper.
- From my notes, drifts should be based on a combination of [O^+] and flag information. In my old matlab routines, there is a 'dmpts' parameter in addition to the drift meter flag which I don't see in the pysat metadata. It's likely my shorthand for something rather than an official name.
- I would also suggest dropping the drifts for 100% O^+.

The first and third are straightforward to implement.  The second requires some more work.

aburrell:
So I've been digging through my C code looking for the different quality controls I used in my dissertation.  Here is the 'RPA is negligible' allowance:

```
/* bad == 0 if there is no sunglint */
 if(ivm_data->qual_flag == 1 && bad == 0)
{
  if(data_ptr->rpa_quality_flag == 3.0 &&
     data_ptr->idens * data_ptr->imassfrac[0] <= 3.0e4)
    {
      /* Test and see if the conditions allow for a data allowance */
      k = 3;

      if(ivm_data->vp_flag < 2 || p_perc < .01) k--;
      if(ivm_data->vm_flag < 2 || m_perc < .01) k--;
      if(ivm_data->vz_flag < 2 || z_perc < .01) k--;

      if(k == 0 && (ivm_data->vp_flag == 2 || ivm_data->vm_flag == 2
		    || ivm_data->vz_flag == 2))
	{
	  printf("%s ADVISEMENT: ram velocity ", module_name);
	  printf("discarded at row [%d/%d]\n", n,
	  	 temp_data->number_samples);
	  data_ptr->vx = 0.0;
	}
      else
	{
	  printf("%s ADVISEMENT: RPA flag triggered at row [%d/%d]\n",
		 module_name, n, temp_data->number_samples);
	  bad = 1;
	}
    }

  if(bad == 0 && data_ptr->idm_quality_num < 9)
    {
      printf("%s ADVISEMENT: IDM flag triggered in row [%d/%d]\n",
	     module_name, n, temp_data->number_samples);
      bad = 1;
    }
}
```

jklenzing:
Currently we have:
- Clean has RPAflag <=1
- Dusty has RPAflag <=3

The drifts in question (if I understand the above code) are those where RPAflag is 3 and ([O+] > 3e4 or X_perc < 0.01)

Should we:
- Open up `clean` to include these drifts?
- Restrict `dusty` to ignore drifts where this is not true?

aburrell:
Well, I used those drifts in all my studies, so I think they are good.  And they are certainly cleaner than the current dusty level.

jklenzing:
OK, as discussed in the meeting, the additional changes based on plasma parameters are now only run for the 'dusty' levels.  Simplifies and streamlines the code.

Also, additional ion temperature cleaning now for 'clean' and 'dusty' levels only.

Ready for new review.

aburrell:
I would test the implementation and create a plot to be sure, but pysatCDF isn't working for me, so I can't load CINDI data :-(

jklenzing:
I will add to the docstring. I don’t think the 100% O+ case is published yet. @rstoneback?

rstoneback:
Nope, no publications on the ultra high O+ case.

Issue #276, opened 2019-09-04
Attributes aren't written to file as expected


rstoneback:
**Describe the bug**
I set some custom attributes to an Instrument object. When I write that object to a netCDF file the attributes are out of date. Values written to files are those set previously by load_netcdf.

**To Reproduce**
1. Load a file, created by pysat, using load_netcdf that has custom attributes
2. Set attributes with the same name directly to Instrument object
3. Write a file using to_netcdf

**Expected behavior**
I expected the custom attributes I set upon the object to end up in the file.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:MacOS
 - 10.14.5

**Additional context**
What is likely happening is the custom attributes are assigned to the metadata object when loaded. I then set new attributes to Instrument. The order of ops in to_netcdf writes Instrument attributes then MetaData attributes. Older attributes in metadata are rewriting values in the netcdf file. We could simply switch the order, but that leads to new bugs.

I expected those custom meta attributes to be transferred to the Instrument object within pysat, automatically. Somewhere in the general load routine. Check into this.

Should also produce a warning, or perhaps even an error, if there are attributes at both Instrument and MetaData level when writing a netCDF4 file.

rstoneback:
There should be a method in .meta that copies attributes from meta over to the instrument object. This should probably be placed in the user facing load routine.

asher-pembroke:
I'm adding the following test to `test_utils.py/TestBasicNetCDF4`

```python
def test_netcdf_meta_change(self):

    self.testInst.load(2009, 1)
    current_units = self.testInst.meta['uts', 'units']

    new_units = current_units + '^2'

    self.testInst.meta['uts', 'units'] = new_units
    
    outfile = os.path.join(self.testInst.files.data_path,
                           'pysat_test_ncdf.nc')

    self.testInst.to_netcdf4(outfile)

    loaded_inst, meta = pysat.utils.load_netcdf4(outfile)

    try:
        assert meta['uts','units'] == new_units
    except AssertionError:
        print('{} != {}'.format(meta['uts','units'], new_units))
        raise
```
I think this test captures the expected behavior.

asher-pembroke:
I'm still a little confused about the metadata vs attribute model and how that differs in the netcdf case. Can someone provide a test that illustrates the problem for me?

asher-pembroke:
I'm getting a different problem from what you describe

```python
testInst = pysat.Instrument(platform='pysat',
                                 name='testing',
[...]

PR #302, opened 2019-10-09
Netcdf meta

asher-pembroke:
# Description

Addresses #276 

Added instrument __setattr__ and __getattr__ so that custom instrument attributes get stored as meta attributes. This fixes a problem where instrument attributes were not being updated properly when writing to netcdf.

In other words,
```python
inst.myattr = 'new'
assert inst.myattr == inst.meta.myattr
assert 'myattr' not in dir(inst)
```
Attributes are set/retrieved by the following priority:
1. set (get)  to (from) `inst._base_attr`
2. set (get) to (from) `inst.meta`
3. set (get) to (from) remaining `inst` attributes

I think this approach is more consistent with the goal of having one place to store instrument metadata, but I would like more input on this design change before going further. 

## Type of change

- Breaking change (see invalidated test below)

# How Has This Been Tested?

- New test added to test_utils.py:
 test_netcdf_attribute_override
- This change _invalidates_ the following test within test_meta.py:
test_transfer_attributes_to_instrument_no_meta_attr

More tests to follow if this approach is acceptable 

**Test Configuration**:
* Mac OS X 10.13.6
* python 3.7.3

# Checklist:

- [ ] New and existing unit tests pass locally with my changes


rstoneback:
The overall feature looks nice. It keeps access across meta and the instrument itself consistent.

The only part I don't really like is having .test_dates written to a file.

jklenzing:
While test_dates aren't particularly useful outside of the testing environment, we should aim for consistency unless there's a good reason not to.  Are these attributes written to operational files?  ie, files to be distributed to the community?

rstoneback:
The tests are failing as the to_netcdf4 routine tries to write the info to files. This will include, say, the ICON IVM files. Yes, it includes community distributed files.

We could switch to a leading _ on test_dates. to_netcdf4 ignores those attributes when writing to a file. Plus, it is consistent with the _ naming of users shouldn't generally mess with it.

jklenzing:
> The tests are failing as the to_netcdf4 routine tries to write the info to files. This will include, say, the ICON IVM files. Yes, it includes community distributed files.
> 
> We could switch to a leading _ on test_dates. to_netcdf4 ignores those attributes when writing to a file. Plus, it is consistent with the _ naming of users shouldn't generally mess with it.

I like switching to to `_test_dates`.  It's an elegant fix, and sets a precedent for future metadata that might be added.

rstoneback:
Great. Asher, please change the test_dates attribute to _test_dates.

Cheers,
Russell

asher-pembroke:
I have changed all appearances of `test_dates` to `_test_dates` and  updated the instrument docs to reflect the change.

jklenzing:
Since we're changing api with the change in `test_date`,  should this be pushed to 3.0?  Adding to agenda.

asher-pembroke:
I don't understand why `pysat/instruments/cosmic_gps.py` still shows as conflicting even though I removed it from my branch. Not sure how to resolve.

jklenzing:
Not sure what's up here.  You should be able to copy over cosmic_gps.py from the latest develop branch and save it here.  I just did a test run to delete a file, commit, restore it, commit.  It shows up as a net-zero change in the final pull requet.

rstoneback:
supplanted by #318

Issue #374, opened 2020-02-11
User Defined MetaData wiped on load


JonathonMSmith:
In version 2.1.0 an instrument with some meta data assigned will lose the user defined meta data after loading data.

1. Install pysat
2. Create Instrument
3. Download data if needed
4. Assign meta data to instrument and check that it was assigned properly
5. Load instrument data
6. Check that meta data is still there
7. See error: AttributeError: No attribute user_meta

**Expected behavior**
I expect the assigned meta data to be unchanged by pysat.

Code test used to produce these results:
```python
from datetime import datetime
import pysat

vefi = pysat.Instrument(platform='cnofs', name='vefi', tag='dc_b')
start = datetime(2010,1,1)
stop = datetime(2010,1,1)
vefi.bounds = (start, stop)
vefi.download(start, stop)
vefi.user_meta = 'somethingMeta'
print('user metadata is : '+vefi.user_meta)

vefi.load(date=start)

print('user metadata is : '+vefi.user_meta)
```
output from pysat version 2.1.0 (paths truncated for space)
```
Downloading data to:  /cnofs/vefi/dc_b/
Attempting to download file for 01 January 2010
Finished.
Updating pysat file list
pysat is searching for cnofs vefi dc_b files.
Found 1 of them.
Updating instrument object bounds.

user metadata is : somethingMeta

Returning cnofs vefi dc_b data for 01 January 2010
Traceback (most recent call last):
  File "/pysat-2.1.0-py3.7.egg/pysat/_instrument.py", line 401, in __getattr__
    return getattr(self.meta, name)
AttributeError: 'Meta' object has no attribute 'user_meta'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "meta_test.py", line 14, in <module>
    print('\nuser metadata is : '+vefi.user_meta+'\n')
  File "/pysat-2.1.0-py3.7.egg/pysat/_instrument.py", line 403, in __getattr__
    raise AttributeError("No attribute {}".format(name))
AttributeError: No attribute user_meta

```
output from pysat version 2.0.0, this is what I expect to happen in version 2.1.0 as well
```
Downloading data to:  /cnofs/vefi/dc_b/
Attempting to download file for 01 January 2010
Finished.
Updating pysat file list
pysat is searching for cnofs vefi dc_b files.
Found 1 of them.
Updating instrument object bounds.

user metadata is : somethingMeta

Returning cnofs vefi dc_b data for 01 January 2010

[...]

PR #375, opened 2020-02-20
Meta fix

JonathonMSmith:
# Description
Small pull request that adds a feature and a unit test.
Now when data is loaded, user defined meta data is transferred to the Meta object from the load before reassigning the loaded meta object to the instrument object.

Currently ssnl fails a lot of unit tests, but this is true of the vanilla develop branch and I think is being addressed in the develop 3 branch.

Addresses #374 

Issue #374 found a bug where user created attributes stored in meta would be lost after loading data. This was due to the fact that meta was replaced completely on load with the meta object from the meta that the instrument specific load function returned.
This fix, which may be clunky or implemented poorly, takes the user defined attributes and transfers them over to the instrument specific meta and then reassigns the instrument meta data similar to the existing behavior.

## Type of change

# How Has This Been Tested?

Issue #374 includes the manual test used to debug, however a unit test has been added that fails using the develop branch, and passes using this branch.

**Test Configuration**:
* CentOS7
* pysat V2.1.0 develop branch and meta_fix branch from my Fork
* Barebones conda envirnoment with only the dependencies for pysat installed
* only tested in python 3.7

# Checklist:

- [X] Make sure you are merging into the ``develop`` (not ``master``) branch
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -1371,6 +1371,13 @@ class Instrument(object):
         else:
             self.data, meta = self._load_data(date=self.date, fid=self._fid)
             if not self.empty:
+                cur_meta = dir(self.meta)
+                new_meta = dir(meta)
+                xtra_meta = [item for item in cur_meta if item not in new_meta]
+                if xtra_meta:
+                    for key in xtra_meta:
+                        xmet_attr = self.meta.__getattribute__(key)
+                        meta.__setattr__(key, xmet_attr)
                 self.meta = meta
 
         # check if load routine actually returns meta
--- a/pysat/tests/test_instrument.py
+++ b/pysat/tests/test_instrument.py
@@ -318,6 +318,11 @@ class TestBasics():
         assert 'pad' in dir(self.testInst)
         assert 'pad' not in dir(self.testInst.meta)
 
+    def test_custom_attribute_retained_on_load(self):
+        self.testInst.my_attr = 'my value'
+        self.testInst.load(2009, 2)
+        assert 'my_attr' in dir(self.testInst.meta)
+
 
     @raises(AttributeError)
     def test_retrieve_bad_attribute(self):
--- a/pysat/tests/test_instrument.py
+++ b/pysat/tests/test_instrument.py
@@ -319,6 +319,7 @@ class TestBasics():
         assert 'pad' not in dir(self.testInst.meta)
 
     def test_custom_attribute_retained_on_load(self):
+        # check attribute remains attached to meta after loading data
         self.testInst.my_attr = 'my value'
         self.testInst.load(2009, 2)
         assert 'my_attr' in dir(self.testInst.meta)


aburrell:
> This was due to the fact that meta was replaced completely on load with the meta object from the meta that the instrument specific load function returned.

The fix you made is a clean way to get rid of this bug without making any major changes.  In this aspect I approve of them.

My concern, which you have also raised, is that we're missing the opportunity to make an improvement.  Namely, not overwriting existing meta data could greatly improve the time it takes to load large amounts of data.

Currently, I think we should get this merged in and open a new issue about the current metadata behaviour.  What do other people think?  The new unit test will ensure that when we clean up the meta data loading, we still have it behave appropriately.

rstoneback:
Thanks for the report and the code. There are a few competing features at work within pysat so we should be careful.

I think this issue actually traces back to something ICON raised. Quick background. When data is loaded, any attributes in the file are attached to metadata. These attributes are then moved to the inst object, see line 1421 in instrument.py. Hard to remember, but they claimed there were some stale attributes and Asher worked on a fix. If I remember correctly, it may have involved moving attributes back to meta when writing a file. 

I'm bringing this up as the unit test is a bit strange for the code fix. In the unit test the test attribute is attached to the inst object, not meta. The change to meta handling shouldn't have anything to do with the test unless attributes are being moved back to meta and then back to inst.

I'll also note that the current code change will not detect a change in content in metadata, leading to a potential stale metadata issue.

We should hold off on merging this until we sort out the life cycle of attributes within pysat.Instrument. Also sounds like a good thing to get documented.

rstoneback:
'not detect a change in content' for metadata parameters that have already have a value.

I'm thinking of C/NOFS here, where the sampling rate changed from 1/2 to 1 second. Other instruments could have configuration changes that require changing the content of variable's metadata but there is no change in name.

rstoneback:
could be issue #302 that is related

rstoneback:
the line self.meta = meta doesn't overwrite data, just points to a new memory location. Filtering the metadata beforehand and copying over will likely take more time.

I think the simplest path forward is to potentially recall #302 (it is directly related) and then add an internal command to the Meta object that prevents new attributes from being set. This command would be run right after the transfer_attributes on line 1421. If a user wants to add an attribute the one place to do it is inst.

rstoneback:
Thanks again for finding this bug Jonathon. 

When you get a chance, can you try out pull #377 and confirm if it sorts out the disappearing attribute issue you are running into?

JonathonMSmith:
Yeah, this seems to work great!

JonathonMSmith:
Since #377 fixes this issue should I close this pull and #374  ?

PR #401, opened 2020-03-23
Netcdf4 export

mddepew:
# Description
Updates to the netcdf4 export function. Improves compatibility with later netCDF4 libraries and provides warnings for unsafe casting of Fill Values in netcdf4 export.

Addresses
NetCDF export fails with current versions of the netCDF4 library with an HDF error. This is because of a conflict between h5py and the hdf libraries from the netCDF4 package. This is resolved by importing netCDF4 in  __init__.py so that the netCDF4 hdf libraries are loaded first.

Currently, the netcdf4 export function casts fillvalues to match the value of the column, but this casting may be unsafe. The most common example is if a nan is being casted to an integer it will become a zero. If zero is a valid value for the column, this will result in the masking of valid data in the exported netCDF file. Added a warning to notify the user if any potentially unsafe casting occurs.

## Type of change

# How Has This Been Tested?
This fix has been tested using the icon_ivm_gsw package and a small subset of ICON L0 data. Ran the zero_to_one processing code, which ends in an export to netCDF4.  Warnings were generated for all variables with unsafe casting for the FillValue.

**Test Configuration**:
* Ubuntu 18.04
* Python 2.7

# Checklist:

### DIFF
--- a/pysat/__init__.py
+++ b/pysat/__init__.py
@@ -82,6 +82,7 @@ else:
     with open(os.path.join(pysat_dir, 'data_path.txt'), 'r') as f:
         data_dir = f.readline()
 
+import netCDF4
 from pandas import Panel, DataFrame, Series, datetime
 from . import utils, model_utils
 from ._constellation import Constellation
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ before_install:
   - pip install 'pandas<0.25' # Temporary for pysat 2.x, remove in 3.0.0
   - pip install 'xarray<0.15' # Temporary for pysat 2.x, remove in 3.0.0
   - pip install pysatCDF >/dev/null
+  # Travis CI currently not working with cftime > 1.1.1
+  - pip install 'cftime==1.1.1'
   # Travis CI currently not working with netCDF4 1.5.3
   - pip install 'netCDF4<1.5.3'
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    - Updates to instrument testing objects for consistency
    - Changed madrigal methods to use `madrigalWeb` as a module rather than
      calling it externally
+   - Added warning when FillValue metadata could lead to unexpected results
+     when writing a netCDF4 file
 - Deprecation Warning
   - custom.add will be renamed custom.attach in pysat 3.0.0
 - Documentation
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -1995,12 +1995,29 @@ class Instrument(object):
         if (coltype == type(' ')) or (coltype == type(u' ')):
             # if isinstance(coltype, str):
             remove = True
+            warnings.warn('FillValue is not an acceptable '
+                          'parameter for strings it will be removed')
+        
         # print('coltype', coltype, remove, type(coltype), )
         if u'_FillValue' in mdata_dict.keys():
             # make sure _FillValue is the same type as the data
             if remove:
                 mdata_dict.pop('_FillValue')
             else:
+                if not np.can_cast(mdata_dict['_FillValue'], coltype):
+                    if 'FieldNam' in mdata_dict:
+                         warnings.warn('FillValue for %s (%s) cannot be safely '
+                                      'casted to %s Casting anyways. '
+                                      'This may result in unexpected behavior'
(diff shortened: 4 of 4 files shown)


mddepew:
Just a note for @rstoneback , the change to the .travis.yml file looks like it is needed across the board for the CI pipeline to run. There are issues with the latest cftime on python 2.7, so locking it at 1.1.1 keeps it working.

rstoneback:
Thanks for sorting that out. We seem to be having a wave of library issues.

mddepew:
Yeah, I think since python 2.7 is EOL a lot of packages haven't really been focused on supporting it anymore.

jklenzing:
Regarding `cftime` on python 3.x, it looks like there is a mismatch between cftime 1.1.1.1 and netCDF4 1.5.2 on Travis CI, possibly all linux distributions.  I don't see the error locally on python 3.7.3 on Mac OS.  From the travis logs:
```
ERROR: Failure: ValueError (numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject)
```
From poking around on stackoverflow, this could be a mismatch caused by different versions of numpy used to make the wheels.  https://stackoverflow.com/questions/25752444/scipy-error-numpy-dtype-size-changed-may-indicate-binary-incompatibility-and

Going to try installing netCDF4 with the --no-binary flag to see if this simplifies things.

jklenzing:
It looks like we might be able to circumvent these issues by going back to a conda install.  Still testing things out.  It's promising, but not working for all versions yet.  Should I write up an issue for this and we call that fix out of scope for this pull?

rstoneback:
Sounds reasonable to me. Thanks for sorting this out.

pysat/__init__.py
```diff
@@ -82,6 +82,7 @@
     with open(os.path.join(pysat_dir, 'data_path.txt'), 'r') as f:
         data_dir = f.readline()
 
+import netCDF4
```
jklenzing, line 85: Why is this imported at the init level?

rstoneback:
add netcdf4 import to __init__.py to ensure that the netCDF4 library is loaded before any other library that would load a conflicting h5py module

On Apr 2, 2020, at 1:22 PM, Jeff Klenzing <notifications@github.com<mailto:notifications@github.com>> wrote:



@jklenzing commented on this pull request.

In pysat/__init__.py<https://nam02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fpysat%2Fpysat%2Fpull%2F401%23discussion_r402520882&data=02%7C01%7Crstoneba%40utdallas.edu%7Cd597ce9ba1374cafe01d08d7d732d561%7C8d281d1d9c4d4bf7b16e032d15de9f6c%7C0%7C1%7C637214485645483199&sdata=d8d4HAHPdkwm7QK3NLisjxezC22vy7jUzntZ94mOfiE%3D&reserved=0>:

> @@ -82,6 +82,7 @@
     with open(os.path.join(pysat_dir, 'data_path.txt'), 'r') as f:
         data_dir = f.readline()

+import netCDF4


Why is this imported at the init level?

pysat/__init__.py
mddepew: Basically to make sure it's loaded before any other package that would load h5py. Newer versions of netCF4 have a custom version of the python HDF5 library that conflicts with h5py.

mddepew: If h5py or a package that imports h5py is loaded before netcdf4 it breaks netcdf4

aburrell: This 'work around' will still break if a packages that uses `h5py` is imported before pysat then?

mddepew: I haven't tested that case, but I would expect so. The real solution is for netCDF4 to resolve their conflict with h5py by using a different namespace or something. In the meantime, we need to either not use newer netCDF4 or use this workaround.

jklenzing: Could this be solved by capping the netCDF4 version in setup.py?

mddepew: I believe so, provided that nothing from the newer netCDF4 libraries is needed.

mddepew: I'm not at the VM that I do testing on right now, but I can see what the latest working version of netCDF4 is

jklenzing: I just remembered that the latest version of netCDF4 is the only version with a wheel for python 3.8.  This could be an issue for any caps.

rstoneback:
Moving forward as the issue is on record.

PR #406, opened 2020-04-08
TST: limits on travis for python 2.7 tests

jklenzing:
# Description

New versions of `pyparsing`, `kiwisolver`, and `beautifulsoup4` have been released which do not support tests on python 2.7.  Limiting these in the travis setup and clarifying the block of pip installs that should not go into the develop-3 version.

Discovered in #405, but opening a separate pull to aid in tracking changes.

## Type of change

- Bug fix -- Testing environment (non-breaking change which fixes an issue)

# How Has This Been Tested?

On Travis CI.

# Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_install:
   # Temporary limits for pysat 2.x, remove in 3.0.0
   - pip install 'pandas<0.25'
   - pip install 'xarray<0.15'
+  - pip install 'pyparsing<3.0.0'
   - pip install 'kiwisolver<1.2.0'
   - pip install 'beautifulsoup4<4.9.0'
   # End of temporary limits
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,12 @@ addons:
 before_install:
   - pip install coveralls
   - pip install future
-  - pip install 'pandas<0.25' # Temporary for pysat 2.x, remove in 3.0.0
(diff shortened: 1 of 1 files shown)


jklenzing:
Changelog already contains "Updates to Travis CI environment"

rstoneback:
What about putting the version's restrictions in setup.py (develop only) to ensure that users working with pysat get the correct versions for our last release with python 2 support?

jklenzing:
We could do that.  Of these three, only beautifulsoup4 is actually a pysat dependency.  The others are dependencies down the chain.

jklenzing:
Looking through the code, this could be another one of those unique Travis issues.  beautifulsoup4 has a dependency `soupsieve`, whose latest version is not supported for python 2.7 according to pypi.  bs4 is configured to impose a cap on soupsieve if python 2 is detected, but Travis insists on installing the latest soupsieve anyway.  This is where the issue comes up in the log.

I am still trying to track down where pyparsing and kiwisolver come in.  We don't require them, and I don't see them in the requirements for bs4 or soupsieve.  From the travis logs, I would suspect that they are coming from bs4 or subprocess32.

rstoneback:
If it is not something that a user would encounter on their own machine then I'm fine with imposing the limits on the travis side. My worry is that at some point down the road someone will need to use python 2 for some reason and I'd like to ensure that pysat v2.2 continues to work, particularly as I may be that person :)

jklenzing:
I don't think I actually hit merge here, but merged a downstream branch, which then closed this out.

PR #412, opened 2020-04-22
ROADMAP: strict_time_flag=True

jklenzing:
# Description

Addresses ROADMAP.

Implements strict_time_flag as True by default.  Removes deprecation warning.

Also, since this results in a ValueError rather than AssertionError for instruments without unique time stamps, adjusts the try / except to catch all errors and provide instrument specific information.

## Type of change

- Breaking change (fix or feature that would cause existing functionality
      to not work as expected)

# How Has This Been Tested?

Via pytest.

# Checklist:


jklenzing:
OPEN QUESTIONS:
- How should pysat treat instruments which violate this rule?  Currently, de2_lang and iss_fpmu do not have unique time stamps, which breaks the end-to-end tests.

jklenzing:
Proposal: We have no control over many instruments, which may not have unique time stamps.  If we impose strict_time_flag = True by default, some instruments (see above) cannot be used.  The code here at most should raise a warning rather than a ValueError, and possibly drop duplicate timestamp values.

https://github.com/pysat/pysat/blob/e8b0983febe86d42b1c8205642f4c715ccbcb663/pysat/_instrument.py#L1409-L1415

aburrell:
Counter: Dropping data is a bad idea, and users need to know if they have duplicate data.  How about changing the ValueError message to say that the data can be loaded by changing the default value of the strict_time flag?  That way users get safe data by default, and any data they want by setting the flag value.

jklenzing:
Since this also technically affects cosmic gps, we need an issue describing what we need here.

jklenzing:
Continue using the error by default, update the unit tests to allow modifying the strict_time_flag to continue testing instruments with non-unique data.

pysat/_instrument.py
```diff
@@ -63,9 +63,9 @@ class Instrument(object):
     temporary_file_list : boolean, optional
         If true, the list of Instrument files will not be written to disk.
         Prevents a race condition when running multiple pysat processes.
-    strict_time_flag : boolean, option (False)
+    strict_time_flag : boolean, option (True)
         If true, pysat will check data to ensure times are unique and
-        monotonic. In future versions, this will be fixed to True.
+        monotonic.
```
aburrell, line 68:
```suggestion
        monotonically increasing.
```

pysat/_instrument.py
```diff
@@ -1406,20 +1406,20 @@ def load(self, yr=None, doy=None, date=None, fname=None, fid=None,
         # ensure data is unique and monotonic
         # check occurs after all the data padding loads, or individual load
         # thus it can potentially check issues with padding or with raw data
-        if (not self.index.is_monotonic_increasing) or (not self.index.is_unique):
+        if not (self.index.is_monotonic_increasing and self.index.is_unique):
+            message = ''
+            if not self.index.is_unique:
+                message = ' '.join((message, 'Loaded data is not unique.'))
+            if not self.index.is_monotonic_increasing:
+                message = ' '.join((message, 'Loaded data is not monotonic',
```
aburrell, line 1414:
```suggestion
                message = ' '.join((message, 'Loaded data is not monotonically',
```

pysat/_instrument.py
```diff
@@ -1406,20 +1406,20 @@ def load(self, yr=None, doy=None, date=None, fname=None, fid=None,
...
+                                           'set inst.strict_time_flag=False',
+                                           'before loading data')))
             else:
-                # warn about changes coming in the future
-                warnings.warn(' '.join(('Strict times will eventually be',
-                                        'enforced upon all instruments.',
-                                        '(strict_time_flag)')),
-                              DeprecationWarning,
-                              stacklevel=2)
+                # warn
```
aburrell, line 1421: Suggest removing uninformative comment.

jklenzing:
Updated as suggested.  Thanks!

PR #439, opened 2020-06-09
BUG: recognize string variables in parse_delimited_files

jklenzing:
# Description

Addresses #337.

`parse_delimited_files` currently assumes all variables must be integers.  Some instruments may use strings.  This improves compatibility with other instrument types to maintain variable type in the routine if integer conversion can't be done.  Adds a `version` variable as a string to the unit tests.

Also changes how this is implemented in the cosmic files.  The existing `shift_uts` variable adds fractional seconds to the time stamp of the file for sorting purposes.  Removed the treatment of GNSS ID as seconds since this is not needed for filenames and is the source of the bug in the issue.

Note that this does not fix potential non-unique timestamps within the data.  A FIXME comment has been added to upgrade to xarray down the line.

Updates for style and PEP8 in cosmic_gps.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

```
import pysat
gps = pysat.Instrument('cosmic', 'gps', 'ionPrf', update_files=True)   
gps.files.files
```
should produce a list of COSMIC files similar to:
```
2018-01-07 00:00:00.000009984    2018.007/ionPrf_C006.2018.007.00.00.G18_2016.1...
2018-12-18 00:00:00.000019968    2018.352/ionPrf_C006.2018.352.00.00.G29_2016.1...
2019-01-01 00:12:00.000619776    2009.001/ionPrf_C001.2009.001.00.12.G26_2013.3...
2019-01-01 00:16:00.000859904    2009.001/ionPrf_C002.2009.001.00.16.G24_2013.3...
2019-01-01 00:17:00.000920064    2009.001/ionPrf_C002.2009.001.00.17.G09_2013.3...
                                                       ...                        
2019-12-31 20:13:00.002629888    2018.365/ionPrf_C006.2018.365.20.13.G15_2016.1...
2019-12-31 20:24:00.004070144    2018.365/ionPrf_C006.2018.365.20.24.G24_2016.1...
2019-12-31 20:40:00.005939968    2018.365/ionPrf_C006.2018.365.20.40.G12_2016.1...
2019-12-31 20:44:00.006500096    2018.365/ionPrf_C006.2018.365.20.44.G02_2016.1...
2019-12-31 20:52:00.007369984    2018.365/ionPrf_C006.2018.365.20.52.G06_2016.1...
Length: 16609, dtype: object
```
Note that the timestamp only goes down to minutes, with less than 1/100th of a second added.

**Test Configuration**:
[...]

Issue #337, opened 2019-12-20
Bug in parse delimited files


rstoneback:
**Describe the bug**
The _files method for dealing with delimited filenames did not work when I tried on COSMIC files.

**To Reproduce**
1. In cosmic_gps.py replace the fixed_width filename parsing call with the delimited version, including the '.' delimiter.
2. method produces an error

**Expected behavior**
Correctly parsed list

**Desktop (please complete the following information):**
 - OS: Max OS 10.12.6

jklenzing:
Can't duplicate this in python 3.7.3 or 3.6.9.

Steps:
1) Replace line 123 of cosmic_gps.py
```
stored = pysat._files.parse_fixed_width_filenames(files, format_str)
```
with
```
stored = pysat._files.parse_delimited_filenames(files, format_str, '.')
```

2) in ipython:
```
import pysat
cosmic = pysat.Instrument(platform='cosmic', name='gps', tag='ionprf')    
cosmic.files.files  
```

Both running the nominal case in `develop` and the modified case above produce the same list of filenames.

jklenzing:
I don't see any other programs like numpy in the routines required here.  Not sure if I'm missing something else.

jklenzing:
Hash ref: 84bcb06

rstoneback:
Thanks for providing an independent check. Either I did something wrong or there is really something a bit wrong with my setup. I’ll double check on my end. On the list.

rstoneback:
I tried this again on the latest develop:

In [5]: gps = pysat.Instrument('cosmic', 'gps', 'ionPrf', update_files=True)
---------------------------------------------------------------------------

ValueError: invalid literal for int() with base 10: 'G26_2013'
> /Users/russellstoneback/Code/pysat/pysat/_files.py(808)parse_delimited_filenames()
    806     # convert to numpy arrays
    807     for key in stored.keys():
--> 808         stored[key] = np.array(stored[key]).astype(int)
    809         if len(stored[key]) == 0:
    810             stored[key] = None

jklenzing:
The issues are stemming from treating the ground station ID as seconds, hence the `second` key being identified as 'G26_2016', a non-integer. If I remove the second assumptions from the cosmic key, the delimited files work fine.

https://cdaac-www.cosmic.ucar.edu/cdaac/cgi_bin/fileFormats.cgi?type=ionPrf

I think our efforts will be better spent on migrating cosmic to xarray for pysat 3.0.0.

[...]

PR #442, opened 2020-06-18
BUG: recognize strings in parse delimit -- pysat 3.0.0 edition

jklenzing:
# Description

Addresses #337.  Twin to #439.

## Type of change

# How Has This Been Tested?

See #439.

# Checklist:

### DIFF
--- a/pysat/instruments/cosmic_gps.py
+++ b/pysat/instruments/cosmic_gps.py
@@ -40,10 +40,10 @@ Warnings
 - Routine was not produced by COSMIC team
 - More recent versions of netCDF4 and numpy limit the casting of some variable
   types into others. This issue could prevent data loading for some variables
-  such as 'MSL_Altitude' in the 'sonprf' and 'wetprf' files. The default UserWarning
-  when this occurs is
-    'UserWarning: WARNING: missing_value not used since it cannot be safely cast
-    to variable data type'
+  such as 'MSL_Altitude' in the 'sonprf' and 'wetprf' files. The default
+  UserWarning when this occurs is
+    'UserWarning: WARNING: missing_value not used since it cannot be safely
+    cast to variable data type'
 
 """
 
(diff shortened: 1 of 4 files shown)


pysat/instruments/cosmic_gps.py
```diff
@@ -113,47 +113,38 @@ def list_files(tag=None, sat_id=None, data_path=None, format_str=None):
...
     # pull of data from the areas identified by format_str
-    stored = pysat._files.parse_fixed_width_filenames(files, format_str)
-
+    stored = pysat._files.parse_delimited_filenames(files, format_str,
+                                                    delimiter='.')
     if len(stored['year']) > 0:
         year = np.array(stored['year'])
         day = np.array(stored['day'])
         hour = np.array(stored['hour'])
         minute = np.array(stored['minute'])
```
aburrell, line 132: So. These are required to be `int`, but `_files.py` no longer requires all key arrays be integers.  Thus, we need a test here to handle the possibility that any one of these four values could be a string.

jklenzing: Hmm.  We do need to add tests for cosmic, since it's outside the eye of pytest.  This does throw an error if I force `minute` as a string:
```
UFuncTypeError: ufunc 'multiply' did not contain a loop with signature matching types (dtype('<U32'), dtype('<U32')) -> dtype('<U32')
```
This should happen if the format string does not match the cosmic files.  We could specifically scan for these and produce a more helpful error message if any of the values are not integer.

Since this part of the code relies on an external list of files, I'm not really sure how to test for this without adding a lot of new infrastructure (say, a good list and a bad list of files stashed in .pysat).

aburrell: > Since this part of the code relies on an external list of files, I'm not really sure how to test for this without adding a lot of new infrastructure (say, a good list and a bad list of files stashed in .pysat).

We will need to develop that to properly unit test the file management.  I think that's outside of this PR, though.  I think that just adding the tests in this routine for now.  Later we can add unit test structure that provides Instruments with bad filenames and ensures the Instrument object handles the failure gracefully.

jklenzing: Poking into this more, if `year` or `month` is passed through as a string, I do not see an error and the index creation line at 142 (new code) works as expected.  The error comes from the calculation of uts at line 133 (new), since we try to multiply a string by an int if `hour` or `minute` is passed through as a string.  I've added a possible testing path forward (for when things go right) as #443, which is likely outside the scope here.  I'm still not sure how to build a test for when things go wrong as suggested above.

jklenzing: The simplest thing I think I can do is to change the four lines above to
```
year = np.array(stored['year']).astype(int)
```
etc, to make sure these are passed through as integers here since they may be strings if something has gone wrong.  It's not a test, but it would clarify the error messages if something is not right.  This allows the routines in `_files` to work generally while keeping things more strict at this specific use case.

pysat/instruments/cosmic_gps.py
```diff
@@ -113,47 +113,38 @@ def list_files(tag=None, sat_id=None, data_path=None, format_str=None):
...
-        # the ground station number in the file encoded as number of seconds
-        second = np.array(stored['second'])
-        # the cosmic satellite number is stored as 0.X, where x is cosmic id
-        uts = hour*3600. + minute*60. + second + np.array(stored['revision'])*1.E-1
-        # do a pre-sort on uts to get files that may conflict with eachother
-        # close together in array order
-        # this ensures that we can make the times all unique
+        uts = hour*3600. + minute*60.
```
aburrell, line 133:
```suggestion
        try:
            uts = hour*3600.0 + minute*60.0
        except (TypeError, UFuncTypeError) as err:
            raise TypeError('unable to construct time from filename\n{:}'.format(err))
```

pysat/instruments/cosmic_gps.py
aburrell: I added a suggestion below.

jklenzing: I'll go with yours.  I think it is more clear if something goes wrong.

PR #441, opened 2020-06-11
MAINT: ICON Instrument updates

jklenzing:
# Description

As ICON prepares to release public data, the instrument support needs to be updated to reflect changes in the file structure.

Starting the changes in the 3.0.0 branch since the instrument tests are more robust here.  Will update the 2.2.0 branch once these are finalized and approved here.

Summary
- Updated instruments to support all publicly available data products for ICON
- Added FTP support for downloads
- Added `methods.icon` to the instrument methods, including data acknowledgement string and download functions
- Use `list_files`  from `methods.general`
- Add xarray support to `utils.load_netcdf`

## Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Files tested and loaded locally, as downloads not yet available.

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``master``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] Add a note to ``CHANGELOG.md``, summarizing the changes

PR #451, opened 2020-06-28
ENH: allow time or Epoch in getitem_xarray

jklenzing:
Alternate method for getitem_xarray.  This should be more compatible and allow either 'time' or 'Epoch' to be used as the key index.  Should be more backwards-compatible for when these are ported to pysat 2.2.

Note this is several commits behind #441.  This allows the test instruments to be maintained for 2.2, though we may want to upgrade in 3.0.

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -433,16 +433,20 @@ class Instrument(object):
             inst[datetime1:datetime1, 'name1':'name2']
 
         """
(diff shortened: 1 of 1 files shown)


rstoneback:
Checking same path forward on setitem.

rstoneback:
Gonna merge since I'm already here :) and ensure your commits go in first. I ended up copying things in on my side.

PR #477, opened 2020-07-22
ICON Demo - Bug Fixes

rstoneback:
# Description

Addresses # (issue)

While working on a demo for working ICON data using pysat I discovered a bug with the orbit iterator, it 'local_time' orbits effectively only worked with instrument reporting near 1 Hz. Further review of the demo indicated several more areas that needed to be changed.

## Type of change

# How Has This Been Tested?

# Checklist:

- [ ] Make sure you are merging into the ``develop`` (not ``master``) branch


pysat/_orbits.py
```diff
@@ -270,7 +276,7 @@ def _equaBreaks(self, orbit_index_period=24.):
             ind = np.hstack((ind, ut_ind))
             ind = np.sort(ind)
             ind = np.unique(ind)
-            logger.info('Time Gap')
+            logger.info('Time Gap at locations ', ut_ind)
```
jklenzing, line 279:
```suggestion
            logger.info(' '.join(('Time Gap at locations', ut_ind)))
```

pysat/_orbits.py
```diff
@@ -220,19 +224,21 @@ def _equaBreaks(self, orbit_index_period=24.):
             # suggest not a true orbit break, but rather bad orbit_index values
             new_ind = []
             for idx in ind:
-                tidx, = np.where(lt_diff[idx - 5:idx + 6] > 0.1)
+                tidx, = np.where(lt_diff[idx - 5:idx + 6] > 10*typical_lt_diff)
```
jklenzing, line 227:
```suggestion
                tidx, = np.where(lt_diff[(idx - 5):(idx + 6)] > 10 * typical_lt_diff)
```

pysat/_orbits.py
```diff
@@ -197,14 +197,18 @@ def _equaBreaks(self, orbit_index_period=24.):
...
+
         # universal time values, from datetime index
         ut_vals = Series(self.sat.index)
         # UT difference
         ut_diff = ut_vals.diff()
 
         # get locations where orbit index derivative is less than 0
         # then do some basic checks on these locations
-        ind, = np.where((lt_diff < -0.1))
+        ind, = np.where((lt_diff < -0.2*typical_lt_diff))
```
jklenzing, line 211:
```suggestion
        ind, = np.where((lt_diff < -0.2 * typical_lt_diff))
```

rstoneback:
It may be worth incorporating some of my modification methods in the demo directly into the cleaning routine.

rstoneback:
This pull includes a couple tweaks to the orbit iterator so they work on data not sampled at 1-Hz. I also included a copy of custom.attach so folks can update their code now in advance of the change coming in 3.0. Also reduces the extent of the pink warning text boxes in the demo. This should be good to go.

pysat/_orbits.py
jklenzing: Just realized this breaks things, since `ut_ind` is not a string.  Working on alternate versions, but the original may be best

jklenzing:
It might be worth doing what we did for other breaking code bits, where `custom.attach` is the code, and `custom.add` just throws a deprecation warning and calls `custom.attach`.  Should smooth things over, while still be easily patched in case we need to fix bugs before 3.0 is released.

aburrell:
YES.  This has been causing me no end of trouble.

rstoneback:
Will do.

rstoneback:
Early requests are in. I also spotted a couple more typos on the nth pass. Hands off on my end until review. Cheers.

rstoneback:
I think you have to open it from within Jupyter? GitHub will also render it if you go to the file in the repo directly.

jklenzing:
Was wondering the same.  If you type `jupyter notebook` in a console window, it opens up the local directory in your web browser, and you can view it there.

aburrell:
Installing jupyter...

aburrell:
So, the `pysat_data_dir = '/Users/russellstoneback/ScienceData'` command doesn't run when you press run.  For people to use the notebooks, I'd recommend replacing "russellstoneback" with "username" or "/Users/russellstoneback" with "$HOME"

rstoneback:
Thank you!

aburrell:
Also note that they need to create the directory.

jklenzing:
image
Should we be including cleaning on the ion density / composition values?

rstoneback:
I’ve visually verified that density and composition are included in cleaning. It is expected for there to good density and comp parameters over a greater range of local times than the other parameters. 

There is likely an issue with composition in the early morning hours but that is (apparently) a limitation of the current flagging on the file generation side.

aburrell:
I keep getting this warning "Notebook ICON_demo.ipynb is not trusted", anyone know why?

rstoneback:
I hadn’t seen that. Maybe MacOS knows you download it from the internet...?

jklenzing:
```
def add_average_local_solar_time(inst):
    """Adds mean profile local solar time as variable """
    # average local time over profile
    inst['average_local_time'] = inst['Local_Solar_Time'].mean(axis=1)
    idx, = np.where(~inst['average_local_time'].isnull())
    inst.data = inst[idx]
    return
```
Local_Solar_Time should be Tangent_Local_Solar_Time

aburrell:
Ah, also say that if you have already set pysat up to skip the data_dir step.  Maybe have the user print it out first.

jklenzing:
oops.  didn't pull latest changes.

aburrell:
I keep getting a TimeoutError on the download.  Gave up on using the Jupyter console (I really have always disliked it) and am running the same commands locally.  Both ways failed.

rstoneback:
I’ve run into that myself and I’m not sure why. It isn’t consistent though.

Most effective I’ve found is to start the download, cancel it with a ctrl-c, then execute the same line again.
I thought it was just issues with my connection. 

I didn’t have this problem earlier, as in when the download code was put in. 
Good to know I’m not the only one, maybe I can adjust how the ftp library connects?

rstoneback:
I added support for ~ and $ variables when setting pysat directory path. I also updated the comments in that area to note the directory must exist, plus only first time pysat users need to set the directory.

aburrell:
Looking over the example plots, a lot of them are very ugly/hard to read.  Some suggestions:

- [ ] For the plots that have the lines going over a blank region, plot with `'.'` instead of using the default
- [ ] Include less subplots.  This is a tutorial, not a "show me all the data!!!!"  Less is more.
- [ ] Set a good example by using the metadata to include units in all the axes and colorbars.

aburrell:
> Good to know I’m not the only one, maybe I can adjust how the ftp library connects?
 Maybe?  This isn't good for a tutorial :P

rstoneback:
I agree having download issues aren’t great for a tutorial but all of my previous downloads went just fine. Whatever the source of the connection trouble it is new.

I’d feel more comfortable using pysftp to connect but I couldn’t get it to work for a standard ftp server.

aburrell:
Repeated code in tutorial
```
# the pysat.Instrument object provides an interface to your file and data needs

# while Instrument objects may be re-used, they are instantiated in each cell in this demo
# to reduce inter-dependence between Jupyter cells
ivm = pysat.Instrument('icon', 'ivm', sat_id='a')
```

aburrell:
Ok, recommend breaking up the 'quick primer' section.  First, print the instrument after download and show how to access list of downloaded files.  Then have a 'load' section, where you print the instrument after loading.  Them show the different ways of loading the variable names in another section.

jklenzing:
Running into some local glitches.  Compiling a list.

I'm also noticing that running the python notebook causes git to log changes locally for me, which I then have to discard before swapping branches.  I would lean toward creating a separate repository / website for the notebooks.

aburrell:
Add a comment about the potential difficulty with downloads due to ftp.  Let me know when all the changes are made and I'll take another look.

rstoneback:
Thanks for all of the feedback. If we are going to put it somewhere else then I can remove the milestone, which helps.

jklenzing:
`pysatTutorials` for the new repository?  Another advantage is that we can push out new / updated tutorials fairly quickly when we get feedback.

I agree with limiting the number of plots / subplots.  If I run these more than once, I start getting feedback that too many plots are open.

jklenzing:
I'll also note that we get a ton of future warnings and deprecations from pandas if using a more recent version.  I think all of these are fixed in 3.0.

rstoneback:
Thanks for the heads up. I’m on 0.24.2 which is the last supported version in the v2.2 setup.py and didn’t see those. I think I’m in the clear on this particular one :)

rstoneback:
> `pysatTutorials` for the new repository? Another advantage is that we can push out new / updated tutorials fairly quickly when we get feedback.
> 
> I agree with limiting the number of plots / subplots. If I run these more than once, I start getting feedback that too many plots are open.

Yeah, pysatTutorials sounds good. I agree it would be much easier to release updates to them outside of the pysat schedule.

I will reduce the number of subplots. I frequently see versions of the MIGHTI plot on team calls, though not by local time, so that is where it started. A team call and a demo aren't the same though.

jklenzing:
I think this is fine, since we still have caps on the pandas versions for this release.  Since I'm using setup.py develop, it doesn't check for the versions when I switch branches, so I get all the pandas 1.0 stuff.

rstoneback:
> Even though we are likely to migrate the ipynb to another repository, we should push the bug updates through, either here or in another pull. Some more thoughts on the tutorial:
> 
> * [ ]  Given the length of time to download data, let's go with two days at most.  Also, MIGHTI data not currently available for the first date, which could be confusing for new users trying to look for Jan 1 data.
> * [ ]  for orbits, maybe two or three plots / subplots at most
> * [ ]  In the plots, label red vs green winds.
> * [ ]  everything @aburrell said

I was just thinking that we should drop the notebook part and push/pull the code changes through so the release of v2.2 may begin.

jklenzing:
sounds good.  Heading over to docs...

PR #478, opened 2020-07-22
STY: test_utils_time -- flake8 and use class structure

jklenzing:
# Description

Updates the test_utils_time file to use a class structure with setup and teardown where appropriate.  Also removes the flake8 exceptions and scrubs code.

## Type of change

# How Has This Been Tested?

locally via pytest --flake8

# Checklist:

### DIFF
--- a/pysat/tests/test_utils_time.py
+++ b/pysat/tests/test_utils_time.py
@@ -59,7 +59,7 @@ class TestParseDate():
             _ = pytime.parse_date('194', '15', '31')
 
 
-def TestCalcFreq():
+class TestCalcFreq():
 
     def setup(self):
         self.year = np.ones(4) * 2001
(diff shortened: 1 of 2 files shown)


jklenzing:
I'm using setup and teardown for the two classes with repeated variables.  I can include empty setup and teardown commands for the other tests if preferred.   I could also rewrite the `TestParseDate` to use parametrize, which might be a little cleaner.  But `TestGetYearDay` and `TestCreateDateRange` both have and require fairly unique variables.  I think this is a little cleaner in terms of readability, but happy to move towards a unified style.

aburrell:
I think we should have a unified style, but that's definitely something to bring up in a group meeting.  And you seem to have a list of options on hand, so it could be a productive discussion!

jklenzing:
Sounds good.  I'll put together some notes for the next meeting.  A lot of my thoughts are still in flux, so I'd be happy to get more input.

PR #479, opened 2020-07-24
DOC: docstring scrub for autodoc build

jklenzing:
This... was more than I expected.

- Scrubs the dosctrings for better imports in sphinx
- Scrubs docstrings for consistency
- fixes broken links

### DIFF
--- a/docs/citing.rst
+++ b/docs/citing.rst
@@ -43,6 +43,6 @@ Citing the publication:
     year      = {2018}
   }
 
-To aid in scientific reproducibility, please include the version number in publications that use this code.  This can be found by invoking `pysat.__version__ `.
+To aid in scientific reproducibility, please include the version number in publications that use this code.  This can be found by invoking ``pysat.__version__ ``.
 
-Information for appropriately acknowledging and citing the different instruments accessed through pysat is sometimes available in the metadata through `inst.meta.info['acknowledgements']` and `inst.meta.info['reference']`.  If this information is missing, please consider improving pysat by either submitting an issue or adding the information yourself.
+Information for appropriately acknowledging and citing the different instruments accessed through pysat is sometimes available in the metadata through ``inst.meta.info['acknowledgements']``` and ``inst.meta.info['reference']``.  If this information is missing, please consider improving pysat by either submitting an issue or adding the information yourself.
--- a/docs/new_instrument.rst
+++ b/docs/new_instrument.rst
@@ -211,6 +211,7 @@ with an Instrument may be updated by adding `update_files=True` at
 instantiation.
 
 .. code:: python
+
    inst = pysat.Instrument(platform=platform, name=name, update_files=True)
 
 The output provided by the list_files function that has been pulled into pysat
@@ -518,8 +519,8 @@ routines for cnofs_vefi and cnofs_ivm for practical uses of the NASA CDAWeb
 support code.
 |br|
 
-.. automodule:: pysat.instruments.methods.nasa_cdaweb
-   :members:
+**Add link here**
+
 
 Madrigal
 --------
@@ -560,5 +561,4 @@ rules of the road.  This can be done by Adding
 to each routine that uses Madrigal data access.
 |br|
 
-.. automodule:: pysat.instruments.methods.madrigal
-  :members:
+**Add link here**
--- a/docs/supported_instruments.rst
+++ b/docs/supported_instruments.rst
@@ -5,14 +5,26 @@ Instrument Templates
 General Instrument
 ------------------
 
-.. automodule:: pysat.instruments.template_instrument
+.. automodule:: pysat.instruments.templates.template_instrument
    :members: __doc__, init, default, load, list_files, list_remote_files, download, clean
+
+Madrigal Pandas
+---------------
+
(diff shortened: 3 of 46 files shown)


jklenzing:
My one open question here is that the module level parameter list
```
Parameters
----------
platform : string
    'cnofs'
name : string
    'ivm'
tag : string
    None supported
sat_id : string
    None supported
```
is translated as this

Screen Shot 2020-07-23 at 8 55 43 PM
Ideas?

jklenzing:
Also, custom functions= docstrings at the module level are weirdly formatted
```
Custom Functions
-----------------
time_shift_to_magnetic_poles : Shift time from bowshock to intersection with
                               one of the magnetic poles
calculate_clock_angle : Calculate the clock angle and IMF mag in the YZ plane
calculate_imf_steadiness : Calculate the IMF steadiness using clock angle and
                           magnitude in the YZ plane
calculate_dayside_reconnection : Calculate the dayside reconnection rate
```
Screen Shot 2020-07-23 at 9 15 55 PM

rstoneback:
While autodoc complains about note vs notes, note makes a better visual impression in the html and pdf. In the pdf the note section goes into a blue highlighted box. Notes gets no additional formatting. There is a similar difference in the html.

jklenzing:
> While autodoc complains about note vs notes, note makes a better visual impression in the html and pdf. In the pdf the note section goes into a blue highlighted box. Notes gets no additional formatting. There is a similar difference in the html.

I see it now.  Fortunately, it's easier to go back than forward.

rstoneback:
I'm poking around the sphinxdoc and RTD documentation and so far I'm not making much progress on how to alter things. I see how to add a logo, but notes vs note, or say how the class members are documented, hasn't come up yet.

rstoneback:
I wonder if the issue with, 
custom functions= docstrings,
is due to 
> time_shift_to_magnetic_poles : Shift time from bowshock to intersection with
                               one of the magnetic poles

being broken over two lines with a bunch of leading space for the second one?

jklenzing:
Screen Shot 2020-07-23 at 9 37 57 PM

That was it.  Hits the 80 character limit though.

jklenzing:
How about this instead?
```
Custom Functions
----------------
time_shift_to_magnetic_poles
    Shift time from bowshock to intersection with one of the magnetic poles
calculate_clock_angle
    Calculate the clock angle and IMF mag in the YZ plane
calculate_imf_steadiness
    Calculate the IMF steadiness using clock angle and magnitude in the YZ plane
calculate_dayside_reconnection
    Calculate the dayside reconnection rate
```
Screen Shot 2020-07-23 at 9 46 18 PM

jklenzing:
So as not to confuse with `custom.attach` we should probably change the header

rstoneback:
Those functions are intended to be used with the custom method queue.

rstoneback:
I looked over numpydoc and it doesn't list parameters as part of a module level docstring. It is possible that I should have made Instrument modules a class rather than just rely on the defacto module behavior of every python file - I'm not putting that change on the table right now though.

rstoneback:
The note section formatting I like appears to be invoking the, 
`.. note:: Note text.`
part of numpydoc and thus isn't really part of the package developers intended standard look of the Notes section of the docstring. They note a caution as it is 'often unnecessary' to use note and doesn't look good in text terminals.  As stated I like the visual difference in the html and pdf but sharing that nugget since I just learned it.

jklenzing:
I'm happy using the note.  It does look better.  As for the parameters, what about
```
Properties
----------
platform
    'omni'
name
    'hro'
tag
    Select time between samples, one of {'1min', '5min'}
sat_id
    None supported
```
Screen Shot 2020-07-23 at 10 31 03 PM

It's a fairly minor shift in terms of edits, and makes the html a bit more readable.

rstoneback:
Cool. 


That is better. There is also an Attributes option that may work.

jklenzing:
Attributes goes a little more in depth.  Same code as above, but replacing Properties with Attributes.
Screen Shot 2020-07-23 at 10 37 15 PM

rstoneback:
I like properties better. Thanks for checking.

jklenzing:
Cool.  I'll put together a fresh pull using Note as the default.

jklenzing:
Also, we had talked about the use of parens around types in docstrings.  As an example, the three types that are employed are:
```
    Parameters
    -----------
    supported_tags : dict
        dict of dicts. Keys are supported tag names for download. Value is
        a dict with 'dir', 'remote_fname', 'local_fname'. Inteded to be
        pre-set with functools.partial then assigned to new instrument code.
    date_array : array_like
        Array of datetimes to download data for. Provided by pysat.
    tag : str or NoneType (None)
        tag or None
    sat_id : (str or NoneType)
        satellite id or None (default=None)
```
Screen Shot 2020-07-23 at 11 00 43 PM

My personal preference is no paren around the type (since html will add another layer of paren), with default value at the end of the description.  Thoughts?

jklenzing:
Most of what we're keeping here is in #480.  Closing this branch.  When we reach consensus on use of parens, I'll start a new pull for that one.

rstoneback:
I’d say no parentheses. I don’t remember putting those in but they are all over the place.

PR #480, opened 2020-07-24
DOC: instrument descriptions

jklenzing:
Going to do this in several smaller review friendly pulls.
- Fixes formatting and broken links in supported_instruments.rst
- Updates top level instrument docstrings for better support in Sphinx

tested via `make html`

### DIFF
--- a/pysat/instruments/templates/template_cdaweb_instrument.py
+++ b/pysat/instruments/templates/template_cdaweb_instrument.py
@@ -22,12 +22,13 @@ tag
 
 Note
 ----
-- Add any notes here
+- Optional section, remove if no notes
 
 
 Warnings
 --------
-- Add any warnings here
+- Optional section, remove if no warnings
+- Two blank lines needed afterward for proper formatting
 
 
 Examples
--- a/pysat/instruments/templates/template_cdaweb_instrument.py
+++ b/pysat/instruments/templates/template_cdaweb_instrument.py
@@ -9,29 +9,37 @@ on the mission, operations, instrumenation, and measurements.
 Also a good place to provide contact information. This text will
 be included in the pysat API documentation.
 
-Parameters
+Properties
 ----------
-platform : string
+platform
     *List platform string here*
-name : string
+name
     *List name string here*
-sat_id : string
+sat_id
     *List supported sat_ids here*
-tag : string
+tag
     *List supported tag strings here*
 
 Note
 ----
-::
+- Add any notes here
 
-    Notes
 
 Warnings
 --------
+- Add any warnings here
+
+
+Examples
+--------
+::
+
+    Example code can go here
 
 
 Authors
 -------
+Author name and institution
 
 """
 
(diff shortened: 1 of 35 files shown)


jklenzing:
Old Style:
Screen Shot 2020-07-23 at 8 55 43 PM

New Style:
Screen Shot 2020-07-23 at 10 31 03 PM

pysat/instruments/templates/template_cdaweb_instrument.py
```diff
@@ -9,29 +9,37 @@
...
 Note
 ----
-::
+- Add any notes here
 
-    Notes
 
 Warnings
 --------
+- Add any warnings here
```
aburrell, line 30: Recommend removing optional categories from template.  But there is good reason to keep them, so maybe we should discuss this.

pysat/instruments/timed_saber.py
```diff
@@ -8,44 +8,42 @@
```
aburrell, line 8: Is this the best place for the "rules of the road"?

pysat/instruments/ucar_tiegcm.py
```diff
@@ -4,21 +4,20 @@
...
-sat_id : string
+sat_id
     None supported
 
+
 Note
 ----
-::
-
     Loads into xarray format.
```
aburrell, line 21: Is this necessary?

pysat/instruments/templates/template_cdaweb_instrument.py
jklenzing: I was thinking of showing how this should be formatted in the templates to help move toward a uniform doc style.  One of the style errors I noticed is that if there are not two blank lines after the warnings, the next line (usually author information) gets added to the warnings box in the html.  What if there was a note that these are optional and should be removed if unused?

pysat/instruments/ucar_tiegcm.py
jklenzing: Probably not, now that we have multiple xarray instruments

pysat/instruments/timed_saber.py
jklenzing: How about as a note?

aburrell:
In rocsat_ivm, recommend removing "Returns" from the "clean" function docstring and replace the "return None"  within that function with "return".  This may fix the coveralls issues in the related pulls and also clean up the docstring.

pysat/instruments/templates/template_cdaweb_instrument.py
aburrell: Sounds good.  Also add that note about the two blank lines.

jklenzing:
Since this pull is pretty big, I was planning on fixing the function level docstrings in a separate pull, including:
- parens in parameter lists
- removal of "returns void" lines

Also, since the ICON instruments have changed significantly, I plan to fix the docstrings for those as part of that pull.

pysat/instruments/timed_saber.py
aburrell: Sure, or maybe reference (not positive that's right either)?  Is this info included in the ackn or ref functions?

jklenzing: ackn and ref are still mostly unused throughout the instruments.  Been working on those as part of the 3.0 branch.

PR #491, opened 2020-07-28
BUG: Updated sdist distribution

rstoneback:
# Description

Addresses # (issue)

The manifest.in files excluded the pysat_testing instruments though those instruments are imported, resulting in an error.
Further, the testing instruments are specifically mentioned in the updated documentation. Finally, downstream modules need this support. This feature is in v3.0 but having it in a release version will/could help smooth the transition to v3.0 for downstream modules.

This pull also includes a minor correction to .readthedocs.yml that was on my system for the rtd_bugfix pull but I missed adding that change as a commit.

I can prepare a valid distribution without this fix but would require the same from anyone else trying to do the same. We should release a v2.2.2. 

## Type of change

# How Has This Been Tested?

I ran `python setup.py sdist` and installed pysat from the resulting distribution, mirroring the content that would be uploaded to PyPI. I verified pysat_testing* files are there and I was able to import pysat. 

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas


rstoneback:
Once more unto the breach, dear friends.

PR #502, opened 2020-08-04
STY: flake8 -- instrument.py

jklenzing:
# Description

- Removes the exceptions for _instrument.py
- renames several local variables to get under the 80 character limit
- Added fix for #504 and new unit test 

## Type of change

# How Has This Been Tested?

via pytest --flake8

# Checklist:

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -383,6 +383,8 @@ class Instrument(object):
 
             # By name
             inst['name']
+            # By list of names
+            inst[['name1', 'name2']]
             # By position
             inst[row_index, 'name']
             # Slicing by row
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -412,8 +412,9 @@ class Instrument(object):
                 try:
                     # integer based indexing
                     return self.data.iloc[key]
-                except TypeError:
+                except (TypeError, ValueError):
                     # If it's not an integer, TypeError is thrown
+                    # If it's a list, ValueError is thrown
                     return self.data[key]
         else:
             return self.__getitem_xarray__(key)
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -2220,13 +2220,13 @@ class Instrument(object):
                     np.array(mdata_dict['FillVal']).astype(coltype)
         return mdata_dict
 
-    def generic_meta_translator(self, meta_to_translate):
+    def generic_meta_translator(self, input_meta):
         '''Translates the metadate contained in an object into a dictionary
         suitable for export.
 
         Parameters
         ----------
-        meta_to_translate : Meta
+        input_meta : Meta
             The metadata object to translate
 
         Returns
@@ -2246,13 +2246,13 @@ class Instrument(object):
         else:
             translation_table = None
         # First Order Data
-        for key in meta_to_translate.data.index:
+        for key in input_meta.data.index:
             if translation_table is None:
-                export_dict[key] = meta_to_translate.data.loc[key].to_dict()
+                export_dict[key] = input_meta.data.loc[key].to_dict()
             else:
                 # Translate each key if a translation is provided
                 export_dict[key] = {}
-                meta_dict = meta_to_translate.data.loc[key].to_dict()
+                meta_dict = input_meta.data.loc[key].to_dict()
                 for orig_key in meta_dict:
                     if orig_key in translation_table:
                         for translated_key in translation_table[orig_key]:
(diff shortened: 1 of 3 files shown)


pysat/_instrument.py
```diff
@@ -1089,31 +1097,29 @@ def __str__(self):
...
         if not self.empty:
-            # if self._fid is not None:
-            #     output_str += 'Filename: ' +
             output_str += 'Date: ' + self.date.strftime('%d %B %Y') + '\n'
             output_str += 'DOY: {:03d}'.format(self.doy) + '\n'
             output_str += 'Time range: '
             output_str += self.index[0].strftime('%d %B %Y %H:%M:%S')
             output_str += ' --- '
-            output_str += self.index[-1].strftime('%d %B %Y %H:%M:%S')+'\n'
+            output_str += self.index[-1].strftime('%d %B %Y %H:%M:%S') + '\n'
```
aburrell, line 1108:
```suggestion
            output_str += self.index[-1].strftime('%d %B %Y %H:%M:%S\n')
```

pysat/_instrument.py
```diff
@@ -1089,31 +1097,29 @@ def __str__(self):
...
                 output_str += self.variables[3 * i + 1].ljust(30)
-                output_str += self.variables[3 * i + 2].ljust(30)+'\n'
+                output_str += self.variables[3 * i + 2].ljust(30) + '\n'
             for i in np.arange(len(self.variables) - 3 * num):
-                output_str += self.variables[i+3*num].ljust(30)
+                output_str += self.variables[i + 3 * num].ljust(30)
             output_str += '\n'
         else:
-            output_str += 'No loaded data.'+'\n'
+            output_str += 'No loaded data.' + '\n'
```
aburrell, line 1122:
```suggestion
            output_str += 'No loaded data.\n'
```

pysat/_instrument.py
```diff
@@ -2710,18 +2723,17 @@ def to_netcdf4(self, fname=None, base_instrument=None, epoch_name='Epoch',
                             for export_units_label in export_units_labels:
                                 new_dict[export_units_label] = \
                                     'Milliseconds since 1970-1-1 00:00:00'
-                            new_dict = self._filter_netcdf4_metadata(new_dict,
-                                                                     coltype,
-                                                                     export_nan=export_nan)
+                            new_dict = self._filter_netcdf4_metadata(
+                                new_dict, coltype, export_nan=export_nan)
```
rstoneback, line 2727: Using the online tool here it looks like the lower line is indented with only 4 spaces. Isn't this supposed to be avoided when extending a statement across multiple lines?

https://www.python.org/dev/peps/pep-0008/#indentation

rstoneback: This comment applies to a number of similar changes to the _filter_netcdf4 calls.

pysat/_instrument.py
```diff
@@ -412,8 +414,9 @@ def __getitem__(self, key):
                 try:
                     # integer based indexing
                     return self.data.iloc[key]
-                except TypeError:
+                except (TypeError, ValueError):
                     # If it's not an integer, TypeError is thrown
+                    # If it's a list, ValueError is thrown
```
jklenzing, line 419: @rstoneback, this should fix #504.  My initial suggestion is on the right track, but pointed tot he wrong spot.

rstoneback: Thanks! I appreciate the update. I won't be able to directly check, in a processing context, until tomorrow at least.

pysat/_instrument.py
jklenzing: The above format is technically correct (at least according to flake8), but may be less clear in readability.  For instance the above is allowed, but the following is not
```
new_dict = self._filter_netcdf4_metadata(new_dict
    coltype, export_nan=export_nan)
```
If I try to allign everything to the paren, it exceeds the 80 character limit.  An alternative would be to use a shorter function name for `_filter_netcdf4_metadata`, or possibly `export_nan`.

pysat/_instrument.py
jklenzing: I've added a unit test to this branch, which is working locally

pysat/_instrument.py
rstoneback: Doesn't having only 4 spaces for the continuation line make it appear like python whitespace formatting?

Would `_filter_nc_meta` make a difference?

I'm going to be out for a bit at a doctor but I'll be back this afternoon.

jklenzing: Still 2 characters too long

jklenzing: Although if we also use a line break, it could work:
```
                            new_dict = \
                                self._filter_nc_meta(new_dict,
                                                     coltype,
                                                     export_nan=export_nan)
```

aburrell: Reading over the link @rstoneback posted, I think the original line is fine.  It looks like this is only an issue when defining a new function, not using a function call.  Also, I believe it's preferable to use the paren whitespace breaks than `\`.

Issue #516, opened 2020-08-14
ENH: Add support for registering multiple instrument modules at once


rstoneback:
**Is your feature request related to a problem? Please describe.**
pysat* packages are organized via pysat*.instruments. ... with multiple instrument modules under pysat*.instruments.
Currently, these individual instruments have to be registered individually.

**Describe the solution you'd like**
Be able to supply pysat*.instruments and have each instrument within registered

**Describe alternatives you've considered**
Current situation.

jklenzing:
Possibilities:
- auto register during installation
- auto register during first import
- ask the user if they want to register all instruments during first import
- only have manual registration (all instruments)
- manual registration (selective instruments)

rstoneback:
To add to the list I was thinking we could have each pysat* package register itself when installed. 

I think the autoimport on first use could be good though perhaps that needs a switch. 
Operational users may prefer that pysat remain static unless otherwise specifically directed. 

We need some support to add a switch, it is an issue.

rstoneback:
The registry module was updated to provide a few easy to use mechanisms for registering one or more instruments at a time. These may be used by pysat dependent packages to register themselves upon installation.

PR #376, opened 2020-02-21
Thread safe

asherp:
# Description

Addresses #304  #309

Using portalocker to avoid potential IO blocking issues. For any IO operation that needs to be thread, processor, or executable-safe, use portalocker's `Lock` and `TemporaryFileLock` context managers. For example, these can be used in conjunction with netcdf's DataSet.

```python
with TemporaryFileLock(fname + '.Lock', pysat.file_timeout) as tfl, \
     netCDF4.Dataset(fname, mode='r', format=file_format) as data:
         pass
```

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for
your test configuration

- Test A
- Test B

**Test Configuration**:
* Mac OS 10.14.6
* python 3.7.3 from miniconda

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``master``) branch
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] Add a note to ``CHANGELOG.md``, summarizing the changes

PR #522, opened 2020-08-28
Registry update

rstoneback:
# Description

Addresses #516 and more

Updates the register package.
 - Changed format of .pysat/user_modules.txt to store packages by platform and name
 - Changed format of pysat.user_modules to be a dict of dicts keyed by platform and name
 - Restricted `pysat_*` instruments to those internal to pysat.instruments
 - Changed use of registry so that only requested module needs to be imported
 - Added support to add multiple packages at once using strings
 - Added support to register an entire instrument package like `pysat.instruments` or `pysatModels.models`
 - Added support to remove all instrument modules from registry by platform name along
 - Added support to remove multiple packages by string platform and name
 - Modified method calls and docstrings
 - Raises error if a module is removed but not registered
 - Raises error if a module is already registered for a given platform and name
 - Expanded unit tests

## Type of change

# How Has This Been Tested?

Unit tests as well as through review of a pysatSpaceWeather instrument, ACE.

# Checklist:

### DIFF
--- a/pysat/__init__.py
+++ b/pysat/__init__.py
@@ -79,20 +79,29 @@ if not os.path.isdir(pysat_dir):
     # user modules file
     with open(os.path.join(pysat_dir, 'user_modules.txt'), 'w') as f:
         f.write('')
-        user_modules = []
+        user_modules = {}
 
 else:
     # load up stored data path
     with open(os.path.join(pysat_dir, 'data_path.txt'), 'r') as f:
         data_dir = f.readline()
+
     # load up stored user modules
-    user_modules = []
+    user_modules = {}
     modules_file = os.path.join(pysat_dir, 'user_modules.txt')
     if os.path.exists(modules_file):
-        with open(modules_file, 'r') as f:
-            for _ in f:
-                if _ != '' and (_ is not None):
-                    user_modules.append(_.strip())
+        with open(modules_file, 'r') as fopen:
+            for line in fopen:
+                if line != '' and (line is not None):
+                    # remove trailing whitespace
+                    line = line.strip()
+                    # stored as platform, name, module string
+                    platform, name, inst_module = line.split(' ')
+                    # dict of dicts, keyed by platform then name
+                    if platform not in user_modules:
+                        user_modules[platform] = {}
+                    # store instrument module string
+                    user_modules[platform][name] = inst_module
     else:
         # write user modules file
         with open(os.path.join(pysat_dir, 'user_modules.txt'), 'w') as f:
--- a/pysat/tests/test_registry.py
+++ b/pysat/tests/test_registry.py
@@ -186,7 +186,7 @@ class TestRegistration():
             # registered has been removed
             for platform, name in zip(self.platforms, self.platform_names):
                 registry.remove(platform, name)
-        except:
+        except Exception:
             # ok if a module has already been removed
             pass
         # ensure things are clean, all have been removed
--- a/pysat/tests/test_registry.py
+++ b/pysat/tests/test_registry.py
@@ -386,7 +386,7 @@ class TestModuleRegistration():
                                module_names]
         self.platforms = ['pysat'] * len(self.platform_names)
         module_strings = ['.'.join((pkg_name, name)) for name in
-                           module_names]
+                          module_names]
         self.modules = [(mod, plat, nam) for mod, plat, nam in
                         zip(module_strings, self.platforms,
                             self.platform_names)]
--- a/pysat/tests/test_registry.py
+++ b/pysat/tests/test_registry.py
@@ -283,9 +283,11 @@ class TestRegistration():
         # verify stored update
         ensure_updated_stored_modules(self.modules)
         # remove them using only platform
-        for platform in np.unique(self.platforms):
+        for i, platform in enumerate(np.unique(self.platforms)):
             registry.remove(platform)
-        # test for removal performed by teardown
+            # doing this one by one ensures more lines tested
+            ensure_not_in_stored_modules([self.modules[i]])
+        # test for removal also performed by teardown
 
         return
 
--- a/pysat/tests/test_registry.py
+++ b/pysat/tests/test_registry.py
@@ -216,3 +216,19 @@ class TestRegistration():
         ensure_updated_stored_modules(self.modules)
 
         return
+
+    def test_platform_removal(self):
+        """Test registering multiple instruments at once"""
+
+        # register all modules at once
+        registry.register(self.module_names)
+        # verify instantiation
+        verify_platform_name_instantiation(self.modules)
+        # verify registration
+        ensure_live_registry_updated(self.modules)
+        # verify stored update
+        ensure_updated_stored_modules(self.modules)
+        # remove them using only platform
+        registry.remove(self.platform_names)
+
+        return
--- a/pysat/tests/test_registry.py
+++ b/pysat/tests/test_registry.py
@@ -191,7 +191,7 @@ class TestRegistration():
             # remove singly to ensure everything that could've been
             # registered has been removed
             for platform, name in zip(self.platforms, self.platform_names):
-                registry.remove(platform, name)
+                registry.remove([platform], [name])
         except ValueError:
             # ok if a module has already been removed
             pass
@@ -271,7 +271,8 @@ class TestRegistration():
         # verify stored update
         ensure_updated_stored_modules(self.modules)
         # remove them using only platform
-        registry.remove(np.unique(self.platforms))
+        uplatforms = np.unique(self.platforms)
+        registry.remove(uplatforms, [None] * len(uplatforms))
         # test for removal performed by teardown
         return
 
@@ -288,7 +289,7 @@ class TestRegistration():
         ensure_updated_stored_modules(self.modules)
         # remove them using only platform
         for i, platform in enumerate(np.unique(self.platforms)):
-            registry.remove(platform)
+            registry.remove([platform], [None])
 
         # doing this one by one ensures more lines tested
 
@@ -308,7 +309,7 @@ class TestRegistration():
         # remove them using only platform
         count = 0
         for platform, name in zip(self.platforms, self.platform_names):
-            registry.remove(platform, name)
+            registry.remove([platform], [name])
             # doing this one by one allows for test that all names
             # for a platform aren't removed
             ensure_not_in_stored_modules([self.modules[count]])
@@ -325,16 +326,7 @@ class TestRegistration():
 
         # remove non-registered modules using only platform
         with pytest.raises(ValueError):
-            registry.remove(['made_up_name'])
-
-        return
-
-    def test_platform_string_removal_error(self):
-        """Test error raised when removing module not present"""
-
-        # remove non-registered modules using only platform
-        with pytest.raises(ValueError):
-            registry.remove('made_up_name')
+            registry.remove(['made_up_name'], [None])
 
         return
 
@@ -365,7 +357,7 @@ class TestRegistration():
             registry.remove(['made_up_name', 'second'], ['made_up_name'])
 
         with pytest.raises(ValueError):
-            registry.remove('made_up_name', ['made_up_name', 'second'])
+            registry.remove(['made_up_name'], ['made_up_name', 'second'])
 
         with pytest.raises(ValueError):
             registry.remove([], ['made_up_name', 'second'])
@@ -374,7 +366,7 @@ class TestRegistration():
             registry.remove([], ['made_up_name'])
 
         with pytest.raises(ValueError):
-            registry.remove([], 'made_up_name')
+            registry.remove([], ['made_up_name'])
 
         return
 
@@ -420,7 +412,7 @@ class TestModuleRegistration():
         # registered
         for plat_name in np.unique(self.platforms):
             try:
-                registry.remove(plat_name)
+                registry.remove([plat_name], [None])
             except ValueError:
                 pass
 
@@ -432,7 +424,7 @@ class TestModuleRegistration():
             # remove singly to ensure everything that could've been
             # registered has been removed
             for platform, name in zip(self.platforms, self.platform_names):
-                registry.remove(platform, name)
+                registry.remove([platform], [name])
         except ValueError:
             # ok if a module has already been removed
             pass
--- a/pysat/tests/test_registry.py
+++ b/pysat/tests/test_registry.py
@@ -196,6 +196,8 @@ class TestRegistration():
         # ensure things are clean, all have been removed
         ensure_not_in_stored_modules(self.modules)
 
+        return
+
     def test_single_registration(self):
         """Test registering package one at a time"""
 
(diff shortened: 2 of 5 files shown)


rstoneback:
To Do:
 - Use pysat testing infrastructure to ensure modules input for registration satisfy all of pysat's minimum requirements
 - Update changelog
 - Update documentation

pysat/utils/registry.py
```diff
@@ -84,48 +111,154 @@ def register(module_name):
...
-    from pysat.utils import registry
+    --------
+    ::
+        from pysat import Instrument, user_modules
+        from pysat.utils import registry
 
-    registry.register('my.package.name.myInstrument')
-    assert 'my.package.name.myInstrument' in user_modules
+        registry.register('my.package.name.myInstrument')
+        assert 'my.package.name.myInstrument' in user_modules
```
rstoneback, line 120: This needs an update.

pysat/utils/registry.py
```diff
@@ -84,48 +111,154 @@ def register(module_name):
...
+        # first, ensure module string directs to something importable
+        try:
+            inst_module = importlib.import_module(module_name)
+        except Exception as error:
+            # log error and then preserve trace and propagate error
+            logger.error(error)
+            raise
+
+        # second, check that module is itself pysat compatible
+
```
rstoneback, line 139: This is where the minimum feature set for a pysat instrument needs to go.

rstoneback: pinging @jklenzing since you are closest to the updated test structure.

rstoneback: A couple options are in the code

rstoneback:
I'll take a pass at documentation once #524 is resolved.

rstoneback:
I also need to pull in the Lock, Network Lock once that is resolved as well in #376 and #525

pysat/__init__.py
```diff
@@ -79,20 +79,29 @@
     # user modules file
     with open(os.path.join(pysat_dir, 'user_modules.txt'), 'w') as f:
         f.write('')
```
aburrell, line 81:
```suggestion
    with open(os.path.join(pysat_dir, 'user_modules.txt'), 'w') as fout:
        fout.write('')
```

pysat/__init__.py
```diff
@@ -79,20 +79,29 @@
...
-        user_modules = []
+        user_modules = {}
 
 else:
     # load up stored data path
     with open(os.path.join(pysat_dir, 'data_path.txt'), 'r') as f:
         data_dir = f.readline()
```
aburrell, line 87:
```suggestion
    with open(os.path.join(pysat_dir, 'data_path.txt'), 'r') as fin:
        data_dir = fin.readline()
```

pysat/__init__.py
```diff
@@ -79,20 +79,29 @@
...
+                    # stored as platform, name, module string
+                    platform, name, inst_module = line.split(' ')
+                    # dict of dicts, keyed by platform then name
+                    if platform not in user_modules:
+                        user_modules[platform] = {}
+                    # store instrument module string
+                    user_modules[platform][name] = inst_module
     else:
         # write user modules file
         with open(os.path.join(pysat_dir, 'user_modules.txt'), 'w') as f:
```
aburrell, line 107: Fix single character variable name here and elsewhere.

pysat/tests/test_registry.py
```diff
@@ -41,53 +44,346 @@ def create_fake_module(full_module_name, platform, name):
...
+
+        # remove non-registered modules using good platform and bad name
+        with pytest.raises(ValueError):
+            registry.remove([self.platforms[0]], ['made_up_name'])
+
+        return
+
+
+class TestModuleRegistration():
+
```
aburrell, line 355: Shouldn't this have a setup and teardown?  I thought we settled on that at the last meeting.

pysat/utils/registry.py
```diff
@@ -15,117 +15,296 @@
...
+    module_names = [module.__name__ + '.' + mod for mod in module_names]
+    # register all of them
+    register(module_names)
 
-    return inst_module
+    return
 
 
-def remove(*module_names):
+def remove(platforms, names=None):
```
aburrell, line 246: Wouldn't it be cleaner to have `names` default to an empty list?  This would make the expected input clearer and not affect the functionality.

pysat/utils/registry.py
```diff
@@ -15,117 +15,296 @@
...
     """Removes module from registered user modules
 
     Parameters
-    -----------
-    module_names : str or multiple str
-        full module package names to remove from registry
-
-    Notes
+    ----------
+    platforms : str of list-like of str
```
aburrell, line 251:
```suggestion
    platforms : str or list-like of str
```

pysat/utils/registry.py
```diff
@@ -15,117 +15,296 @@
...
+        One or more platform identifiers to remove
+    names : str of list-like of str
+        Name identifiers, paired with platforms, to remove.
+        If names is None, then all instruments under
+        `platforms` will be removed. Supports a mixed
+        combination of name labels and None. (default=None)
```
aburrell, line 257:
```suggestion
    names : str or list-like of str
        Name identifiers, paired with platforms, to remove.
        If the names element paired with the platform element is None,
        then all instruments under the specified platform will be
        removed. Should be the same type as `platforms`. An empty list
        will default to removing all instruments for all specified
        platforms. (default=[])
```

pysat/utils/registry.py
aburrell: I provided a suggestion in the docstring, but didn't make changes elsewhere.  It's really just the `is None` evaluation line that would need to be changed to `len(names) == 0` and also add a ValueError check for `len(names) != len(platforms)`.  This should probably be done anyhow.

rstoneback: General guidance I've read is that objects that are modifiable in place shouldn't be used as defaults in the method signature. If names=[], or names='hi there!', is in the call signature and names gets modified in-place within the method, those changes will persist between calls since the signature definition code is only run at definition.

while it seems contrary to the whole point of the feature, I've seen the best-practices method signature is
```
def new_method(var1, keyword=None):
    if keyword is None:
        keyword = []
```

pysat/__init__.py
rstoneback: Those changes were in #376, and now #525, and in progress.

pysat/tests/test_registry.py
rstoneback: Fixed.

pysat/utils/registry.py
rstoneback: I adopted the suggestion apart from the default.

pysat/utils/registry.py
rstoneback: If there are no other opinions I prefer the second option.

 Also thinking we can write a method in utils, something like verify_pysat_standard (compatibility?). Input a instrument module, get a report on how it fares against pysat standards, and test against methods. Basically a simple way to test things in development.

pysat/utils/registry.py
aburrell: >  objects that are modifiable in place shouldn't be used as defaults in the method signature

Yes, I am ok with `names` being an arg instead of a kwarg.  Is there any instance where someone wouldn't be expected to enter both?  It's required in every other area.

General guidance also recommends avoiding the use of `isinstance` for testing your input.  You can get around this by using `np.asarray` on the input.  This will be more robust in instances where someone has an input that for some reason isn't interpreted as a string (as has happened before and will likely happen again).

pysat/utils/registry.py
aburrell: > A couple options are in the code

It may be more useful for discussion to put them in the conversation here as examples or put them on a slide/wiki page as examples to be discussed at the next meeting.

pysat/utils/registry.py
rstoneback: >> Is there any instance where someone wouldn't be expected to enter both? It's required in every other area.

If a user wants to remove all instruments for a given platform(s). I like having a minimal number of required inputs but I am ok with changing names to an arg. I'll have something tomorrow.

 I do like to be able to support a single string or list-like of strings with the same code. I've tried using np.asarray and it hasn't worked out for me so far.
```
In [19]: for item in np.asarray('hello'): 
    ...:     print(item) 
    ...:                                                                                                                                                              
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-bff22eb1af27> in <module>
----> 1 for item in np.asarray('hello'):
      2     print(item)
      3 

TypeError: iteration over a 0-d array

```

I've tried duck typing multiple times on the string vs list issue. I've also tried checking for iter or other things like that. None are quite right so far.

pysat/utils/registry.py
rstoneback: Good point.

prep work in containing method.
```
inst_module = importlib.import_module(module_name)      
validate = itc.InstTestClass()
```

First option, using the test class as is.
```
        class Foo(object):
            pass
        # parse string to get package part and instrument module part
        parse = module_name.split('.')

        validate.package = Foo()
        validate.package.__name__ = '.'.join(parse[:-1])
        validate.test_modules_standard(parse[-1])
        validate.test_standard_function_presence(parse[-1])
```
No modifications to the testing code is required. Requires a dummy class and relation of prep code to use by validate isn't clear unless familiar with validate.

Using the test class with a slight mod for outside use. Doesn't impact use of current test code.
```
        validate.test_modules_standard('', inst_module)
        validate.test_standard_function_presence('', inst_module)
```
I like the direct input of the module under test, though it comes with the downside of having a '' leading input and changes to input of test methods.

rstoneback: I went ahead and cleaned things up in favor of the first option and using the code as it exists. Not worth adding in an optional code path for something there is already have a solution for. I spent a bit of time on the comments instead.

pysat/utils/registry.py
rstoneback: Switched to list-like for args platforms and names. Similar change to other methods that allowed a string or list like of strings, now just list-like.

pysat/tests/test_registry.py
```diff
@@ -1,9 +1,12 @@
...
 """
+
+import numpy as np
 import pytest
 import sys
 
+import pysat
 from pysat import user_modules
 from pysat.utils import registry
 from pysat import Instrument
```
aburrell, line 12: Since you're now importing pysat, is it really useful to separately importing `user_modules` and `Instrument`?

pysat/utils/registry.py
```diff
@@ -15,117 +15,305 @@
...
 
-    registry.register('my.package.name.myInstrument')
-    assert 'my.package.name.myInstrument' in user_modules
+    """
+
+    for module_name in module_names:
+        # first, ensure module string directs to something importable
+        try:
+            inst_module = importlib.import_module(module_name)
+        except Exception:
```
aburrell, line 143: Do we know what type of Exception is expected?

pysat/utils/registry.py
aburrell: Cool.  For future reference, this was an alternative:
```
platforms = np.asarray(platforms)

if platforms.shape == ():
    platforms = np.array([platforms])
```

rstoneback: Cool. I figured the registry wasn't a high user interaction area so probably not worth the potential future maintenance costs of fancy inputs. I'll remember this for areas that will see more use though!

pysat/utils/registry.py
rstoneback: My intent with this try: except is verify we can import the supplied input and provide some context to any error if one occurs. The raise a couple lines below preserves the original error and stack trace for the user to deal with.  

No errors are allowed on an import to ensure, to the extent possible, that the registry only has good packages written to it.

rstoneback: There use to be more of these in earlier versions. It started with my logging the error in a variety of places, since, logging is good. But most of the entries would end up like,
```
pysat ERROR: 'platname1'
```
which wasn't all that helpful. I tried a few versions but non seemed to really fit the bill. This one was a residual. After thinking about my response above, it really doesn't do very much.

rstoneback: I also have a similar modification over in _instrument.py, so worth sorting out our style.
```
                # not a native pysat.Instrument
                try:
                    mod = user_modules[self.platform][self.name]
                    inst = importlib.import_module(mod)
                except Exception:
                    estr = ' '.join(('pysat was either unable to locate the',
                                     'module in the registry or',
                                     'successfully import the module',
                                     'for platform:', self.platform,
                                     'and name:', self.name))
                    logger.error(estr)
                    raise

```

rstoneback: The notion of friendlier error message came up on a recent review that threw a KeyError when I forgot a tag.

aburrell: Well, the reason I don't like this is that there are too many things it will catch.  I like catching exceptions and giving friendlier messages, but specifying which one would be more useful.  For example, you could have:
```
try:
    mod = user_modules[self.platform][self.name]
    inst = importlib.import_module(mod)
except KeyError as kerr:
    estr = "unknown platform or name supplied to user modules: {:} or {:} not in {:}".format(self.platform, self.name, user_modules.__repr__())
    logger.error(estr)
    raise KeyError(kerr)
except ImportError as ierr:
    estr = "unable to locate or import module for platform {:}, name {:}".format(self.platform, self.name)
    logger.error(estr)
    raise ImportError(ierr)
```

Then, if something else breaks we'll still get it in the intended form and can fix or catch it as appropriate.

rstoneback:
Review comments have been incorporated. Ready for review @aburrell. Cheers.

pysat/tests/test_registry.py
```diff
@@ -36,58 +40,385 @@ def create_fake_module(full_module_name, platform, name):
...
+        self.platform_names = [snip.split('pysat')[-1][1:] for snip in
+                               module_names]
+        self.platforms = ['pysat'] * len(self.platform_names)
+        module_strings = ['.'.join((pkg_name, name)) for name in
+                          module_names]
+        self.modules = [(mod, plat, nam) for mod, plat, nam in
+                        zip(module_strings, self.platforms,
+                            self.platform_names)]
+
+        # remove any existing support which may be let over
```
aburrell, line 390:
```suggestion
        # remove any existing support which may be left over
```
?

pysat/tests/test_registry.py
```diff
@@ -36,58 +40,385 @@ def create_fake_module(full_module_name, platform, name):
...
+        # errors if you try to remove a package that isn't
+        # registered
+        for plat_name in np.unique(self.platforms):
+            try:
+                registry.remove([plat_name], [None])
+            except ValueError:
+                pass
```
aburrell, line 397: Testing may benefit from raising a warning for platforms that can't be removed.

rstoneback: The ValueError was raised if someone tried to remove a platform that wasn't registered. I removed that error.

 logger.info now gets a message that the package wasn't registered.

rstoneback:
I modified the pull based upon conversation on Wed. Errors raised are reduced. Added an overwrite flag to register. 
ValueError raised by .register if someone tries to register a module that has a platform and name combo already registered for a different module, and overwrite set to False. Same module is ok. Existing registry may be updated with overwrite set to True.

aburrell (approved):

Issue #495, opened 2020-07-29
DOC: Update and Confirm Release Checklist


rstoneback:
**Is your feature request related to a problem? Please describe.**
Bugs in pysat have have resulted directly from the undocumented release process.

**Describe the solution you'd like**
A checklist for release, along with the potential for multiple 'pull request' templates.

**Describe alternatives you've considered**
Current ad hoc process. Could also have an automated release process but that isn't likely to catch all things that need checking. 

**Additional context**
checklists are used in life or death situations so supporting an improved software release should be no problem

rstoneback:
Current checklist has a draft on the wiki:
https://github.com/pysat/pysat/wiki/Checklist-for-Release

aburrell:
```
sphinx-build -E -b html docs dist/docs
sphinx-build -b linkcheck docs dist/docs
```

For doc tests in travis.yml

PR #551, opened 2020-09-17
Travis docs

aburrell:
# Description

Updated travis to include separate doc testing and updated PR template to improve release pull requests.

Addresses #495 

## Type of change

# How Has This Been Tested?

This needs to be tested by running on Travis

**Test Configuration**:
* Operating system
* Version number

# Checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,8 +26,8 @@ your test configuration
 - Test B
 
 **Test Configuration**:
-* Operating system
-* Version number
+* Operating system: iOs
+* Version number: Python 3.X
 * Any details about your local setup that are relevant
 
 # Checklist:
@@ -42,3 +42,6 @@ your test configuration
 - [ ] New and existing unit tests pass locally with my changes
 - [ ] Any dependent changes have been merged and published in downstream modules
 - [ ] Add a note to ``CHANGELOG.md``, summarizing the changes
+
+If this is a release PR, replace the first item of the above checklist with the release
+checklist on the wiki: https://github.com/pysat/pysat/wiki/Checklist-for-Release
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,7 +26,7 @@ your test configuration
 - Test B
 
 **Test Configuration**:
-* Operating system: iOs
+* Operating system: Hal
 * Version number: Python 3.X
 * Any details about your local setup that are relevant
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
   - conda activate test-environment
   - while read requirement; do conda install --yes $requirement; done < requirements.txt
   # Get dependencies not provided by conda from pip
-  - pip install coveralls sphinx-build pytest-flake8
+  - pip install coveralls sphinx pytest-flake8
   # set up data directory
   - mkdir /home/travis/build/pysatData
   # install pysat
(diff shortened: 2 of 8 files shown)


jklenzing:
While we're updating this, we might consider dropping the `-vs` option to cut down on the log size.  Running a test on this now:  https://travis-ci.org/github/pysat/pysat/jobs/728055037

aburrell:
> While we're updating this, we might consider dropping the -vs option to cut down on the log size. Running a test on this now

I'm onboard with that, I don't think we're losing anything important since the failures will still pop up.

.travis.yml
```diff
@@ -27,19 +37,19 @@ install:
...
-# command to run tests
-script:
- - pytest -vs --flake8 --cov=pysat/
-
 after_success:
  - coveralls --rcfile=setup.cfg
+
+allow_failures:
+  - name: 'docs'
+    python: '3.6'
```
jklenzing, line 55: I think this needs to be part of the jobs 
matrix above.  See examples at https://docs.travis-ci.com/user/customizing-the-build/

.travis.yml
```diff
@@ -2,8 +2,18 @@ language: python
 dist: xenial
 matrix:
```
jklenzing, line 3: `jobs` is replacing `matrix` at Travis CI
```suggestion
jobs:
```

.github/pull_request_template.md
```diff
@@ -26,8 +26,8 @@ your test configuration
 - Test B
 
 **Test Configuration**:
-* Operating system
-* Version number
+* Operating system: iOs
```
jklenzing, line 29: Do we want to use iOs as the example?

jklenzing (commented):
A few changes based on the config report at https://travis-ci.org/github/pysat/pysat/builds/728077458/config
Looks like some of this may have been addressed already.  Couldn't add a comment at the top of the yml file, but travis wants an os line (os: linux) in the file as well.

We could also take this opportunity to add NEP 029 compliance.  Pushed a test branch with this and the proposed fixes to see how it works.

Do we want to move flake8 testing to the latest version?  The config now would theoretically be more stable, but this could let us know about changes down the pipeline.

.github/pull_request_template.md
aburrell: I chose it because it's not a real operating system, but it's clearly an operating system.  Open to other suggestions.  LCARS perhaps?

jklenzing: lol.  I was largely thinking of if we were implying support for a system we don't support.  We could always go with a fictional example.

aburrell: Go ahead and pick your fave, Battlestar Galactica or Star Trek?

.travis.yml
aburrell: I didn't want to mess with your matrix, but yeah.  I'll switch to jobs and change this.

.github/pull_request_template.md
```diff
@@ -26,7 +26,7 @@ your test configuration
 - Test B
 
 **Test Configuration**:
-* Operating system: iOs
+* Operating system: Hal
```
jklenzing, line 29: :thumbsup:

aburrell:
I wanted the flake8 on a stable version.  Given the bug that popped up when I started running python 3.8, I think we should add the new versions as quickly as possible, but keep them as allowed failures if there initially issues until those problems are fixed.

Made the suggested changes, fingers crossed!

jklenzing:
The travis files are all in various states of compliance throughout the pysat codes.  I think the example in pysatModels is closest to what we want  in general.

jklenzing (approved):
Travis config is working.  Thanks!

rstoneback:
Thanks @aburrell and @jklenzing!

PR #604, opened 2020-11-23
Add timezone

aburrell:
# Description

Implements aware timestamps throughout the core code, will address #514 

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

# How Has This Been Tested?

This has been tested by running the new unit tests in `test_utils_time`.  Stated running through old unit tests and `test_instrument` fails.  Not sure if this is due to the current way file information is saved?

I see the benefits of using aware datetime objects, especially with the SpaceWeather instruments, but most people I know don't use them.  Do we want to require this as input?  How lax/strict should we be?

```______________________________________ ERROR at setup of TestBasics.test_basic_instrument_load _______________________________________

self = <pysat.tests.test_instrument.TestBasics object at 0x10a567fd0>

    def setup(self):
        reload(pysat.instruments.pysat_testing)
        """Runs before every method to create a clean testing setup."""
        self.testInst = pysat.Instrument(platform='pysat', name='testing',
                                         num_samples=10,
                                         clean_level='clean',
>                                        update_files=True)

pysat/tests/test_instrument.py:32: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pysat/_instrument.py:377: in __init__
    ignore_empty_files=ignore_empty_files)
pysat/_files.py:179: in __init__
    self.refresh()
pysat/_files.py:394: in refresh
    self._store()
pysat/_files.py:300: in _store
    if stored_files.eq(self.files).all():
../../../Library/Python/3.7/lib/python/site-packages/pandas-0.24.2-py3.7-macosx-10.14-x86_64.egg/pandas/core/ops.py:1870: in flex_wrapper
    return self._binop(other, op, level=level, fill_value=fill_value)
../../../Library/Python/3.7/lib/python/site-packages/pandas-0.24.2-py3.7-macosx-10.14-x86_64.egg/pandas/core/series.py:2452: in _binop
    copy=False)
../../../Library/Python/3.7/lib/python/site-packages/pandas-0.24.2-py3.7-macosx-10.14-x86_64.egg/pandas/core/series.py:3664: in align
    broadcast_axis=broadcast_axis)
[...]

PR #611, opened 2020-12-03
Updates to timezone branch

jklenzing:
Updates to #604 

Fixes all unit tests except netcdf-related tests in test_utils.

The run time has increased dramatically (~4 hrs on my local machine).  Travis CI errors in the midst of the orbit tests due to run time issues.  @aburrell, any ideas?

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -1890,7 +1890,11 @@ class Instrument(object):
         else:
             # Specify the dimension, if not otherwise specified
             if 'dim' not in kwargs:
-                kwargs['dim'] = self.index.name
+                if 'time' in self.data.indexes:
+                    kwargs['dim'] = 'time'
+                elif 'Epoch' in self.data.indexes:
+                    kwargs['dim'] = 'Epoch'
+
             concat_func = xr.concat
 
         # Assign the concatonated data to the instrument
--- a/pysat/instruments/pysat_testing_xarray.py
+++ b/pysat/instruments/pysat_testing_xarray.py
@@ -123,16 +123,16 @@ def load(fnames, tag=None, inst_id=None, sim_multi_file_right=False,
     else:
         root_date = dt.datetime(2009, 1, 1, tzinfo=dt.timezone.utc)
 
+    index = index.tolist()
     if malformed_index:
-        index = index.tolist()
         # nonmonotonic
         index[0:3], index[3:6] = index[3:6], index[0:3]
         # non unique
         index[6:9] = [index[6]] * 3
 
-    data = xarray.Dataset({'uts': ((epoch_name), index.tolist())},
-                          coords={epoch_name: index.tolist()})
-    
+    data = xarray.Dataset({'uts': ((epoch_name), index)},
+                          coords={epoch_name: index})
+
     # need to create simple orbits here. Have start of first orbit
     # at 2009,1, 0 UT. 14.84 orbits per day
     time_delta = dates[0] - root_date
--- a/pysat/tests/test_custom.py
+++ b/pysat/tests/test_custom.py
@@ -198,10 +198,10 @@ class TestBasics():
 
         self.testInst.custom.attach(custom1, 'add')
         self.testInst.load(2009, 1)
-        assert (self.testInst.data['doubleMLT'] == 2.0
-                * self.testInst['mlt']).all()
-        assert (self.testInst.data['tripleMLT'] == 3.0
-                * self.testInst['mlt']).all()
+        assert (self.testInst.data['doubleMLT'].values
+                == 2.0 * self.testInst['mlt'].values).all()
+        assert (self.testInst.data['tripleMLT'].values
+                == 3.0 * self.testInst['mlt'].values).all()
         assert len([kk for kk in self.testInst.data.keys()]) == self.ncols + 2
 
     def test_add_dataframe_w_meta(self):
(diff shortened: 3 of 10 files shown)


jklenzing:
Running a quick check via timeit:
```
import timeit
timeit.timeit("import datetime as dt; dt.datetime(2009, 1, 1)", number=10000)
timeit.timeit("import datetime as dt; dt.datetime(2009, 1, 1, tzinfo=dt.timezone.utc)", number=10000)
```
yields 2.6 ms for the naive datetime, 6.9 ms for the aware one.

pysat/_instrument.py
```diff
@@ -1890,7 +1890,11 @@ def concat_data(self, new_data, prepend=False, **kwargs):
         else:
             # Specify the dimension, if not otherwise specified
             if 'dim' not in kwargs:
-                kwargs['dim'] = self.index.name
+                if 'time' in self.data.indexes:
+                    kwargs['dim'] = 'time'
+                elif 'Epoch' in self.data.indexes:
+                    kwargs['dim'] = 'Epoch'
+
```
jklenzing, line 1897: Following conventions elsewhere in _instrument.  The change of the behaviour of `index` in the test instruments broke this

pysat/instruments/pysat_testing_xarray.py
```diff
@@ -123,16 +123,16 @@ def load(fnames, tag=None, inst_id=None, sim_multi_file_right=False,
     else:
         root_date = dt.datetime(2009, 1, 1, tzinfo=dt.timezone.utc)
 
+    index = index.tolist()
     if malformed_index:
-        index = index.tolist()
```
jklenzing, line 127: moving list outside for consistent behaviour

pysat/tests/test_custom.py
```diff
@@ -198,10 +198,10 @@ def custom1(inst):
...
         self.testInst.custom.attach(custom1, 'add')
         self.testInst.load(2009, 1)
-        assert (self.testInst.data['doubleMLT'] == 2.0
-                * self.testInst['mlt']).all()
-        assert (self.testInst.data['tripleMLT'] == 3.0
-                * self.testInst['mlt']).all()
+        assert (self.testInst.data['doubleMLT'].values
+                == 2.0 * self.testInst['mlt'].values).all()
+        assert (self.testInst.data['tripleMLT'].values
+                == 3.0 * self.testInst['mlt'].values).all()
```
jklenzing, line 204: The xarray object will store the new data under a different dimension, hence the error.

pysat/tests/test_instrument.py
```diff
@@ -842,17 +842,17 @@ def test_custom_instrument_load_3(self):
     # -------------------------------------------------------------------------
     def test_basic_data_access_by_name(self):
         self.testInst.load(self.ref_time.year, self.ref_doy)
-        assert np.all(self.testInst['uts'] == self.testInst.data['uts'])
+        assert np.all(self.testInst['mlt'] == self.testInst.data['mlt'])
```
jklenzing, line 845: Comparing for data values makes more sense here and below, as for some tests the 'uts' will be retrieved as different data types now that timestamps are utc.

pysat/tests/test_instrument.py
```diff
@@ -1179,7 +1181,7 @@ def test_list_comprehension(self):
         assert dates.is_monotonic_increasing
 
         # dates are unique
-        assert np.all(np.unique(dates) == dates.values)
+        assert len(np.unique(dates)) == len(insts)
```
jklenzing, line 1184: This fits the above comment better.  Each date in the list should be unique.  Was breaking due to different timestamp types

pysat/tests/test_instrument.py
```diff
@@ -1962,7 +1964,6 @@ def test_set_bounds_fname_with_frequency_and_width(self):
         out = pds.date_range(start_date, stop_date - pds.DateOffset(days=1),
                              freq='2D').tolist()
         # convert filenames in list to a date
-        date_list = []
```
jklenzing, line 1965: not used

jklenzing (commented):
rationale for some of the non-obvious changes

aburrell:
> adding timezone info increases the runtime for calling a datetime object by a factor of 3.6.

This is potentially a deal breaker.  And in my opinion tripling the runtime is a deal breaker.  

Possible paths (not all mutually exclusive):
1) keep naive datetime, assuming it is UTC, and fix Instrument today/yesterday/tomorrow
2) allow aware datetime input, using an internal routine to convert it to UTC and then recast it as naive
3) just be slow and use aware datetimes everywhere

I think (1) and (2) may be enough, since it makes sure time is handled consistently internally and allows users to do whatever they want with time when interfacing with pysat.  Since we are only working with one time zone, (3) can be considered overkill.

jklenzing:
I'm leaning in the same direction.  I am up for 1 and 2, though I am somewhat concerned that 2 might have the same runtime issues if we scan every date.  Unless this is only for the download / load routines.

aburrell:
It would only be for load and download, since those are the only outside initialization points.

rstoneback:
Sounds good to me.

On Dec 3, 2020, at 11:06 AM, Angeline Burrell <notifications@github.com<mailto:notifications@github.com>> wrote:



It would only be for load and download, since those are the only outside initialization points.

rstoneback:
the Instrument class has a method called _filter_datetime_input that already checks for input dates with hours, minutes, and seconds in a few places. It may be sufficient to work on that method.

jklenzing:
Should we close these two branches and update the issue then?

aburrell:
Let's leave them here until the new branch exists so the existing treatment isn't lost.

aburrell:
Closed with merge of #622

PR #622, opened 2020-12-10
Added soft UTC testing

aburrell:
# Description

As discussed in #604 and #611 it is too time intensive to define timezones to all timestamps.  However, we should not accidentally treat data from other timezones as UTC.  This PR addresses #514 by:

1) using `utcnow` for today/yesterday/tomorrow,
2) using Instrument time filtering function to convert times with timezones to UTC and then converting them to naive objects in UTC time, and
3) changing pds.DateOffset to dt.timedelta when appropriate.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Added a new unit test.  Also did a time analysis of pandas vs datetime incrementation:

```
import timeit

timeit.timeit("import datetime as dt; import pandas as pds; dt.datetime.utcnow() + dt.timedelta(days=1)", number=10000)
timeit.timeit("import datetime as dt; import pandas as pds; dt.datetime.utcnow() + pds.DateOffset(days=1)", number=10000)
```

I got 0.026728390992502682 from the datetime call and 0.7501618139940547 from the pandas call.

**Test Configuration**:
* Operating system: OS X Mojave
* Version number: 3.7

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``main``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] Add a note to ``CHANGELOG.md``, summarizing the changes

PR #624, opened 2020-12-11
STY: Use datetime.timedelta, not pandas.DateOffset

jklenzing:
# Description

Replaces `pds.DateOffset` with `dt.timedelta` throughout.   The exceptions are where keywords of `months` or `years` are required, which are not supported by `timedelta`.  This was generated after the results of runtime tests from #622.  There is no significant difference in run time between this and the develop-3 branch, but putting this up as a draft for discussion.

This branch:  https://travis-ci.com/github/pysat/pysat/builds/207865614
develop3:  https://travis-ci.com/github/pysat/pysat/builds/207863988

The test `test_str_w_padding` in test_instrument's output was updated to reflect the change from pandas.

## Type of change

# How Has This Been Tested?

Tested via pytest.

**Test Configuration**:
* Mac OS X 10.15.7
* Python 3.8.2

# Checklist:

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -307,13 +307,14 @@ class Instrument(object):
             self.multi_file_day = multi_file_day
 
         # Initialize the padding
-        if isinstance(pad, dt.timedelta) or pad is None:
+        if isinstance(pad, (dt.timedelta, pds.DateOffset)) or pad is None:
             self.pad = pad
         elif isinstance(pad, dict):
-            self.pad = dt.timedelta(**pad)
+            self.pad = pds.DateOffset(**pad)
         else:
-            raise ValueError(''.join(['pad must be a dict, NoneType, or ',
-                                      'pandas.DateOffset instance.']))
+            raise ValueError(' '.join(['pad must be a dict, NoneType,',
+                                       'datetime.timedelta, or',
+                                       'pandas.DateOffset instance.']))
 
         # Store kwargs, passed to standard routines first
         self.kwargs = {}
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -307,10 +307,10 @@ class Instrument(object):
             self.multi_file_day = multi_file_day
 
         # Initialize the padding
-        if isinstance(pad, pds.DateOffset) or pad is None:
+        if isinstance(pad, dt.timedelta) or pad is None:
             self.pad = pad
         elif isinstance(pad, dict):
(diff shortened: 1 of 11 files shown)


pysat/_instrument.py
```diff
@@ -307,10 +307,10 @@ def __init__(self, platform=None, name=None, tag=None, inst_id=None,
...
         # Initialize the padding
-        if isinstance(pad, pds.DateOffset) or pad is None:
+        if isinstance(pad, dt.timedelta) or pad is None:
             self.pad = pad
         elif isinstance(pad, dict):
-            self.pad = pds.DateOffset(**pad)
+            self.pad = dt.timedelta(**pad)
         else:
             raise ValueError(''.join(['pad must be a dict, NoneType, or ',
                                       'pandas.DateOffset instance.']))
```
aburrell, line 316: This would need to be updated.

aburrell (commented):
I believe there are a few instances were pandas DateOffset specific functionality is used, but I don't recall where.

pysat/tests/test_instrument.py
```diff
@@ -1209,17 +1209,15 @@ def test_right_bounds_with_next(self):
     def test_set_bounds_with_frequency(self):
         """Test setting bounds with non-default step"""
         start = self.ref_time
-        stop = self.ref_time + pds.DateOffset(days=14)
-        stop = stop.to_pydatetime()
+        stop = self.ref_time + dt.timedelta(days=14)
```
jklenzing, line 1212: This is the only bit I found where a specific pandas input was required, but since we're immediately converting to datetime we can just contract the code.

pysat/_instrument.py
jklenzing: I'm not sure if we want to change to datetime here.  We would lose the possibility of padding data by months or years (potetnially useful in simple datasets like f107?).  Thoughts?

aburrell: I think that padding is useful and warrants keeping pds.DateOffset here.  Should we allow timedelta as well, though?

jklenzing: How about we allow both, and if it's a dict, assume pds.DateOffset to interpret, since it is more robust?

aburrell: > assume pds.DateOffset to interpret

We may need an additional test to enforce this, but yes.

jklenzing: updated the test using parametrize to ensure that both options work

aburrell (approved):

Issue #697, opened 2021-02-17
Deprecated non-testing Instruments


aburrell:
**Is your feature request related to a problem? Please describe.**
Real instruments that live in pysat are deprecated in pysat 3.0

**Describe the solution you'd like**
Add a deprecation warning, a note in the docstring, and a unit test for the deprecation warning.

**Describe alternatives you've considered**
Surprise the users!

- [x] champ_star
- [x] cnofs_ivm
- [x] cnofs_plp
- [x] cnofs_vefi
- [x] cosmic_gps
- [x] de2_lang
- [x] de2_nacs
- [x] de2_rpa
- [x] de2_wats
- [x] demeter_iap
- [x] dmsp_ivm
- [x] icon_euv
- [x] icon_fuv
- [x] icon_ivm
- [x] icon_mighti
- [x] iss_fpmu
- [x] jro_isr
- [x] omni_hro
- [x] rocsat1_ivm
- [x] sport_ivm
- [x] superdarn_grdex
- [x] supermag_magnetometer
- [x] sw_dst
- [x] sw_f107
- [x] sw_kp
- [x] timed_saber
- [x] timed_see
- [x] ucar_tiegcm

aburrell:
I have now finished all instrument that are not pysatNASA.  I'll leave the rest to @jklenzing and @JonathonMSmith.

PR #744, opened 2021-03-18
DEP: pysatNASA instruments

jklenzing:
# Description

Partially addresses #697 by adding deprecation warnings to all Instrument modules and methods associated with the pysatNASA package.

## Type of change

# How Has This Been Tested?

Added unit tests for all currently tested routines.
# Checklist:

- [ n/a] My changes generate no new warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
         omni_hro, rocsat1_ivm, sport_ivm, timed_saber, timed_see
    - The usage of a numeric string for `sat_id` to specify number of points
      in test instruments
-   - SpaceWeather, Incubator (DEMETER), and Madrigal instrument method
+   - SpaceWeather, Incubator (DEMETER), Madrigal, cdaweb, and icon instrument
+     methods
+   - The de2 and icon constellation objects
 - Documentation
    - Updated docstrings with deprecation notes
 - Bug Fix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    - Updated docstrings with deprecation notes
 - Bug Fix
    - Closes files after download in NASA CDAWeb methods
+   - Updated the ICON constellation object to reflect new instrument standards
 
 ## [2.2.2] - 2020-11-23
 - New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
       - the pysatMadrigal: jro_isr, dmsp_ivm
       - the pysatSpaceWeather: sw_dst, sw_f107, sw_kp
       - the pysatModels: ucar_tiegcm
+      - the pysatNASA: cnofs_ivm, cnofs_plp, cnofs_vefi, de2_lang, de2_nacs,
+        de2_rpa, de2_wats, icon_euv, icon_fuv, icon_ivm, icon_might, iss_fpmu,
+        omni_hro, rocsat1_ivm, sport_ivm, timed_saber, timed_see
    - The usage of a numeric string for `sat_id` to specify number of points
      in test instruments
    - SpaceWeather, Incubator (DEMETER), and Madrigal instrument method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
         omni_hro, rocsat1_ivm, sport_ivm, timed_saber, timed_see
    - The usage of a numeric string for `sat_id` to specify number of points
      in test instruments
-   - SpaceWeather, Incubator (DEMETER), Madrigal, cdaweb, and icon instrument
+   - SpaceWeather, Incubator (DEMETER), Madrigal, CDAWeb, and ICON instrument
      methods
-   - The de2 and icon constellation objects
+   - The DE2 and ICON Constellation objects
 - Documentation
    - Updated docstrings with deprecation notes
 - Bug Fix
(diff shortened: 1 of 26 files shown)


aburrell (changes requested):
The routines in the instrument sub-modules that are not fed into the Instrument class need deprecation warnings (e.g., omni_hro) and then any that have unit tests need the warnings tested (not bothering to add deprecation tests to those who aren't already covered).  

Additionally, in my package pulls I deprecated the relevant pysat.instrument.methods subpackages.  Are you planning on doing that here or in a separate pull request?

jklenzing:
@aburrell, the only thing left is the deprecation warnings on the constellations.  I can't run the code locally to verify, travis has it still failing.  Any thoughts?

CHANGELOG.md
```diff
@@ -23,9 +23,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
       - the pysatMadrigal: jro_isr, dmsp_ivm
       - the pysatSpaceWeather: sw_dst, sw_f107, sw_kp
       - the pysatModels: ucar_tiegcm
+      - the pysatNASA: cnofs_ivm, cnofs_plp, cnofs_vefi, de2_lang, de2_nacs,
+        de2_rpa, de2_wats, icon_euv, icon_fuv, icon_ivm, icon_might, iss_fpmu,
+        omni_hro, rocsat1_ivm, sport_ivm, timed_saber, timed_see
    - The usage of a numeric string for `sat_id` to specify number of points
      in test instruments
-   - SpaceWeather, Incubator (DEMETER), and Madrigal instrument method
+   - SpaceWeather, Incubator (DEMETER), Madrigal, cdaweb, and icon instrument
```
aburrell, line 31:
```suggestion
   - SpaceWeather, Incubator (DEMETER), Madrigal, CDAWeb, and ICON instrument
```

CHANGELOG.md
```diff
@@ -23,9 +23,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+     methods
+   - The de2 and icon constellation objects
```
aburrell, line 33:
```suggestion
   - The DE2 and ICON Constellation objects
```

aburrell (commented):
That's because the warnings are in the module, which isn't run.  There's no way to get them to be raised.  The Instrument warnings were raised after the init method was run.  But there's no equivalent for the Constellation 😞

pysat/constellations/de2.py
```diff
@@ -1,8 +1,23 @@
...
+
+import pysat
+
+
+warnings.warn(" ".join(["This constellation",
+                        "has been removed from the pysat-managed",
+                        "Instruments in pysat 3.0.0, and now resides in",
+                        "pysatNASA:",
+                        "https://github.com/pysat/pysatNASA"]),
+              DeprecationWarning, stacklevel=2)
```
aburrell, line 20:
```suggestion 
import pysat
```

jklenzing:
> That's because the warnings are in the module, which isn't run. There's no way to get them to be raised. The Instrument warnings were raised after the init method was run. But there's no equivalent for the Constellation 😞

Should we just ignore the constellation module?  If someone does try to run it, they should get the deprecation warnings from the instruments themselves.

pysat/constellations/de2.py
aburrell: Here and ICON

aburrell (commented):
I think you can solve the problem by adding a warning to the constellation object that triggers if the DE2 or ICON constellations are used as input.  Similar to what you did for the unit test.

pysat/constellations/icon.py
```diff
@@ -1,8 +1,23 @@
...
+
+import pysat
+
+
+warnings.warn(" ".join(["This constellation",
+                        "has been removed from the pysat-managed",
+                        "Instruments in pysat 3.0.0, and now resides in",
+                        "pysatNASA:",
+                        "https://github.com/pysat/pysatNASA"]),
+              DeprecationWarning, stacklevel=2)
```
jklenzing, line 20:
```suggestion
```

aburrell: For future reference, you can edit my suggestion.

jklenzing:
🤞

jklenzing:
Update: The ICON constellation object does not work, and has not for a while.  I've documented this in pysat/pysatNASA#49 to decide on future standards.  Looking through old versions, this would have been broken in python 2.2.0.

aburrell:
> Update: The ICON constellation object does not work, and has not for a while. I've documented this in pysat/pysatNASA#49 to decide on future standards. Looking through old versions, this would have been broken in python 2.2.0.

Can it be minimally fixed here?

jklenzing:
I've pushed a simple version through here.  If that works, I'll try to put together a fixed version.

jklenzing:
Update: this is not as bad as I thought.  Pushing through a version with 8 instruments.  The other 4 are the component winds that go into the vector winds products, so i think we can leave these off.  In a practical sense, I suspect anyone wanting to run the constellation will probably just build a custom one with the bits they want to compare.

jklenzing:
Another difference that may need to be addressed here: pysatNASA uses the CDAWeb archive for ICON, pysat 2.x uses the Berkeley FTP server.  This results in a slight difference in file names (mixed case, underscores), which means that the data path migration system will not work for users who migrate from 2 to 3.  We should probably include this in the deprecation warning here. 

The alternative is to do the server switch here, which would basically introduce a breaking change in 2.3.

aburrell:
Just add that info to the deprecation warning.  It'd be good to add a "migration guide" page to the pysatNASA docs.

jklenzing:
proposed addition to warnings: 
"Note that the ICON files will be retrieved from the CDAWeb server in pysatNASA, please see the migration guide there for more details."

We can add a link once I get the docs up there, or just leave it like this.

aburrell:
Counterproposal: "Note that the ICON files are retrieved from the Berkley FTP server here and from the CDAWeb server in pysatNASA, resulting in a difference in local file names.  Please see the migration guide there for more details."

jklenzing:
"Coverage decreased (-6.0e-05%)" 😅

aburrell (approved):
Nice revisions, they all look good to me!

Issue #764, opened 2021-04-09
Common Constellation operations


aburrell:
**Is your feature request related to a problem? Please describe.**
When I try to use a Constellation, I find it to be not very useful, because there's not common time index and I can't specify a custom function that operates on the Constellation instruments as a group.  At the moment, it's basically just a list of Instruments.

**Describe the solution you'd like**
- [x] Common time index for a Constellation
- [x] Constellation-level custom functions that use data from one Instrument to operate on another, and so on.
- [x] `empty` property at constellation level
- [x] `const_to_inst` method that will output the constellation data in an Instrument object.

**Describe alternatives you've considered**
Just not use Constellation objects

**Additional context**
Constellations are not as robustly tested as Instruments in a workflow context, due to their development history.  These are just the first two things that I thought should exist.

PR #785, opened 2021-05-17
Added useful Constellation attributes

aburrell:
# Description

Added constellation methods and attributes needed to make it possible to use the Constellation object in data analysis.

Partially Addresses #764 

## Type of change

# How Has This Been Tested?

Added unit tests.  Also:

```
from pysatSpaceWeather.instruments import ace_mag, ace_swepam
import pandas as pds

half_window = pds.DateOffset(minutes=10)
imf_tag = 'realtime'

imf = pysat._constellation.Constellation(instruments=[pysat.Instrument(inst_module=ace_mag, tag=imf_tag, clean_level='clean', pad=half_window), pysat.Instrument(inst_module=ace_swepam, tag=imf_tag, clean_level='clean', pad=half_window)])
print(imf)

pysat Constellation object:
---------------------------
2 Instruments

Index Platform Name Tag Inst_ID
-------------------------------
0 'ace' 'mag' 'realtime' ''
1 'ace' 'swepam' 'realtime' ''


Loaded Data Statistics
----------------------
No loaded data.
```

```
imf.download(start=imf.today())
imf.load(date=imf.today())
print(imf)
pysat Constellation object:
---------------------------
2 Instruments

Index Platform Name Tag Inst_ID
-------------------------------
0 'ace' 'mag' 'realtime' ''
1 'ace' 'swepam' 'realtime' ''


Loaded Data Statistics
----------------------
Date: 17 May 2021
Common Time range: 17 May 2021 15:03:00 --- 17 May 2021 16:59:59
Number of variables: 15

Variable Names:
jd_ace_mag_realtime     sec_ace_mag_realtime    status_ace_mag_realtime 
                                  ...                                   
sw_proton_dens          sw_bulk_speed           sw_ion_temp 
```

**Test Configuration**:
* Version number: Python 3.7
* Any details about your local setup that are relevant: RC of pysatSpaceWeather

# Checklist:

### DIFF
--- a/pysat/_constellation.py
+++ b/pysat/_constellation.py
@@ -165,13 +165,11 @@ class Constellation(object):
     # -----------------------------------------------------------------------
     # Define all hidden methods
 
-    def _empty(self, data=None, all_inst=True):
+    def _empty(self, all_inst=True):
         """Boolean flag reflecting lack of data
 
         Parameters
         ----------
-        data : NoneType, pds.DataFrame, or xr.Dataset
-            Data object
         all_inst : bool
             Require all instruments to have data for the emtpy flag to be
             False, if True.  If False, the empty flag will be False if any
@@ -186,7 +184,9 @@ class Constellation(object):
         """
         eflags = [inst.empty for inst in self.instruments]
 
-        if all_inst:
+        if len(eflags) == 0:
+            eflag = True
+        elif all_inst:
             eflag = np.any(eflags)
         else:
             eflag = np.all(eflags)
@@ -218,7 +218,7 @@ class Constellation(object):
                     # If desired, determine the resolution
                     if self.index_res is None:
                         if inst.index.freq is None:
-                            out_res = (inst.index[1:]-inst.index[:-1]).mean()
+                            out_res = (inst.index[1:] - inst.index[:-1]).mean()
                         else:
                             out_res = pds.to_datetime(1, units=inst.index.freq)
                 else:
@@ -237,7 +237,7 @@ class Constellation(object):
                     # If desired, determine the resolution
                     if self.index_res is None:
                         if inst.index.freq is None:
-                            new_res = (inst.index[1:]-inst.index[:-1]).mean()
+                            new_res = (inst.index[1:] - inst.index[:-1]).mean()
                         else:
                             new_res = pds.to_datetime(1, units=inst.index.freq)
 
--- a/pysat/_constellation.py
+++ b/pysat/_constellation.py
@@ -147,7 +147,7 @@ class Constellation(object):
         output_str += '----------------------\n'
         if not self.empty:
             output_str += 'Date: ' + self.date.strftime('%d %B %Y') + '\n'
-            output_str += '{:s}Time range: '.format(
+            output_str += '{:s}time range: '.format(
                 "Common " if self.common_index else "Full ")
             output_str += self.index[0].strftime('%d %B %Y %H:%M:%S')
             output_str += ' --- '
--- a/pysat/_constellation.py
+++ b/pysat/_constellation.py
@@ -171,7 +171,7 @@ class Constellation(object):
         Parameters
         ----------
         all_inst : bool
-            Require all instruments to have data for the emtpy flag to be
+            Require all instruments to have data for the empty flag to be
             False, if True.  If False, the empty flag will be False if any
             instrument has data. (default=True)
 
--- a/pysat/_constellation.py
+++ b/pysat/_constellation.py
@@ -3,10 +3,20 @@
 # Full author list can be found in .zenodo.json file
 # DOI:10.5281/zenodo.1199703
 # ----------------------------------------------------------------------------
-""" Created as part of a Spring 2018 UTDesign project.
+""" Class for Instrument constellations
+
+Note
+----
+Created as part of a Spring 2018 UTDesign project.
+Updated by AGB, May 2021, NRL
+
 """
 
+import datetime as dt
 import numpy as np
+import pandas as pds
+
+from pysat import utils
 
 
 class Constellation(object):
@@ -14,10 +24,17 @@ class Constellation(object):
 
     Parameters
     ----------
-    const_module : string
-        Name of a pysat constellation module
-    instruments : list-like
+    const_module : str or NoneType
+        Name of a pysat constellation module (default=None)
+    instruments : list-like or NoneType
         A list of pysat Instruments to include in the Constellation
+        (default=None)
+    index_res : float or NoneType
+        Output index resolution in seconds or None to determine from
+        Constellation instruments (default=None)
+    common_index : bool
+        True to include times where all instruments have data, False to
+        use the maximum time range from the Constellation (default=True)
 
     Attributes
     ----------
@@ -33,7 +50,8 @@ class Constellation(object):
     # -----------------------------------------------------------------------
     # Define the magic methods
 
-    def __init__(self, const_module=None, instruments=None):
+    def __init__(self, const_module=None, instruments=None, index_res=None,
+                 common_index=True):
         """
         Constructs a Constellation given a list of instruments or the name of
         a file with a pre-defined constellation.
@@ -41,9 +59,16 @@ class Constellation(object):
         Parameters
         ----------
         const_module : string
-            Name of a pysat constellation module
+            Name of a pysat constellation module (default=None)
         instruments : list-like
             A list of pysat Instruments to include in the Constellation
+            (default=None)
+        index_res : float or NoneType
+            Output index resolution in seconds or None to determine from
+            Constellation instruments (default=None)
+        common_index : bool
+            True to include times where all instruments have data, False to
+            use the maximum time range from the Constellation (default=True)
 
         Raises
         ------
@@ -72,6 +97,10 @@ class Constellation(object):
 
             self.instruments.extend(list(instruments))
 
+        # Set the index attributes
+        self.index_res = index_res
+        self.common_index = common_index
+
         return
 
     def __getitem__(self, *args, **kwargs):
@@ -88,7 +117,9 @@ class Constellation(object):
         """
 
         out_str = "".join(["pysat.Constellation(instruments=",
-                           "{:})".format(self.instruments)])
+                           "{:}, index_res=".format(self.instruments),
+                           self.index_res.__repr__(), ", common_index=",
+                           self.common_index.__repr__(), ")"])
         return out_str
 
     def __str__(self):
@@ -96,10 +127,11 @@ class Constellation(object):
 
         """
 
+        ninst = len(self.instruments)
+
         output_str = 'pysat Constellation object:\n'
         output_str += '---------------------------\n'
-
-        ninst = len(self.instruments)
+        output_str += "{:d} Instruments\n".format(ninst)
 
         if ninst > 0:
             output_str += "\nIndex Platform Name Tag Inst_ID\n"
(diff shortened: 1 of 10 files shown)


aburrell:
No changes to the docs needed, because there are no constellation examples.

aburrell:
Increases in coverage require a change to pysat test Instruments or test Constellations that are a bit complicated.

pysat/_constellation.py
```diff
@@ -108,10 +140,119 @@ def __str__(self):
...
+            for inst in self.instruments:
+                if stime is None:
+                    # Initialize the start and stop time
+                    stime = inst.index[0]
+                    etime = inst.index[-1]
+
+                    # If desired, determine the resolution
+                    if self.index_res is None:
+                        if inst.index.freq is None:
+                            out_res = (inst.index[1:] - inst.index[:-1]).mean()
```
rstoneback, line 221: What if an instrument say
- switched sample rate during the loaded data
- had a variety of gaps which would increase the mean past the actual sample rate 

I noticed there is a `calc_freq` in utils.time based off the minimum time difference. I think you wrote that. Would that work here?

pysat/_constellation.py
```diff
@@ -108,10 +140,119 @@ def __str__(self):
...
+                    else:
+                        if stime > inst.index[0]:
+                            stime = inst.index[0]
+                        if etime < inst.index[-1]:
+                            etime = inst.index[-1]
+
+                    # If desired, determine the resolution
+                    if self.index_res is None:
+                        if inst.index.freq is None:
+                            new_res = (inst.index[1:] - inst.index[:-1]).mean()
```
rstoneback, line 240: Same as above.

pysat/_constellation.py
```diff
@@ -136,6 +277,62 @@ def bounds(self, value=None):
 
         return
 
+    @property
+    def date(self):
+        """Date for loaded data."""
+        if len(self.index) > 0:
+            return utils.time.filter_datetime_input(self.index[0])
```
rstoneback, line 284: I double checked in pysat and it sets the date in the same way, using first index. At times I've thought maybe a midpoint of loaded data or something else may be better. Checking in to see if you have any additional thoughts on the matter. If not, consider it resolved.

pysat/_constellation.py
```diff
@@ -136,6 +277,62 @@ def bounds(self, value=None):
...
+        return self._empty(all_inst=True)
+
+    @property
+    def index(self):
+        """Returns time index of loaded data."""
+        return self._index()
+
+    def today(self):
+        """Returns UTC date for today, see pysat.Instrument for details."""
+        return utils.time.today()
```
rstoneback, line 302: I'm guessing if you thought `tomorrow` and `yesterday` should also be implemented in `utils.time` you would have. I am curious, why `today` and not the others?

pysat/_instrument.py
```diff
@@ -1880,19 +1878,9 @@ def bounds(self, value=None):
...
-        else:
-            if 'time' in self.data.indexes:
-                return len(self.data.indexes['time']) == 0
-            elif 'Epoch' in self.data.indexes:
-                return len(self.data.indexes['Epoch']) == 0
-            else:
-                return True
+        """Boolean flag reflecting lack of data, True if there is no data.
+        """
+        return self._empty()
```
rstoneback, line 1883: Thanks for catching the duplicated lines.

rstoneback (commented):
Thanks @aburrell for improving constellation. A few questions below.

pysat/_constellation.py
aburrell: Probably! 🦊  I will go take a look at that.

pysat/_constellation.py
aburrell: No, I think the date is fine as the first time.  Since load assumes by default you're loading 1+ days, having the date as the date of the first time is a good representation of the loaded data.

pysat/_constellation.py
aburrell: For the written record, because `tomorrow` and `yesterday` are simple additions and `today` is the time call that actually does much work.  Also, grabbing `today` from Instrument didn't work (caused `pysat` to not import 🦇 ) and that motivated the move to `utils.time`.

pysat/_constellation.py
aburrell: Added two new functions to fix this and other problems I noted when going back and looking at this more closely!  Will re-request a review after I finish the unit tests.

aburrell:
Tests failing because of pysatSpaceWeather release 😛

pysat/utils/time.py
```diff
@@ -124,6 +157,46 @@ def calc_freq(index):
...
+    res_sec : float
+       Resolution value in seconds
+
+    See Also
+    --------
+    pds.offsets.DateOffset
+
+    Reference
+    ---------
+    Separating alpha and numberic portions of strings, as described in:
```
jklenzing, line 179:
```suggestion
    Separating alpha and numeric portions of strings, as described in:
```

jklenzing (approved):
Looks good.  Errors are fixed in #797

rstoneback:
Ran into a download issue with pysatNASA https://github.com/pysat/pysatNASA/pull/69 while checking this out.

pysat/_constellation.py
```diff
@@ -136,6 +279,62 @@ def bounds(self, value=None):
...
+    def date(self):
+        """Date for loaded data."""
+        if len(self.index) > 0:
+            return utils.time.filter_datetime_input(self.index[0])
+        else:
+            return None
+
+    @property
+    def empty(self):
+        """Boolean flag reflecting lack of data, True if there is no Instrument
```
rstoneback, line 292: I actually posted a couple incorrect suggested changes since I didn't properly get this at first. I tried the code before reading the docstring 🐒

Perhaps `empty` isn't the best word here since you are really checking if it is full (all_inst=True).

I don't think there will ever be good data for both ivm 'a' and 'b' in the ICON constellation so that one would always be empty.

pysat/_constellation.py
```diff
@@ -108,10 +140,121 @@ def __str__(self):
...
                     i, inst.platform, inst.name, inst.tag, inst.inst_id)
         else:
-            output_str += "No loaded Instruments\n"
+            output_str += "No assigned Instruments\n"
+
+        # Display loaded data
+        output_str += '\n\nLoaded Data Statistics\n'
+        output_str += '----------------------\n'
+        if not self.empty:
+            output_str += 'Date: ' + self.date.strftime('%d %B %Y') + '\n'
```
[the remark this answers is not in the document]
rstoneback: This actually popped up when I had a mod in _empty while still behind. That being said, as in the pull, print(icon) says there is no loaded data when there is definitely some loaded data.

My guess is that `empty` is driven by requirements on the common index but for print I think we should provide feedback on the data it does have.

pysat/_constellation.py
```diff
@@ -108,10 +140,121 @@ def __str__(self):
...
+            eflag = True
+        elif all_inst:
+            eflag = np.any(eflags)
+        else:
+            eflag = np.all(eflags)
+
+        return eflag
+
+    def _index(self):
+        """Returns a common time index for the loaded data
```
rstoneback, line 197: For something like ICON which is never `full` I'd still like to be able to use the common index. I get that if the index covers times where everything has data, a single instrument without means no common area. But again, ICON.

Brainstorming idea:
What if there was a strict option? If strict, the do current behavior. If not strict, index covers common range of times for instruments with data.

I think there could be/are problems with my suggestion. I'll keep thinking myself.

pysat/utils/time.py
```diff
@@ -73,45 +74,77 @@ def parse_date(str_yr, str_mo, str_day, str_hr='0', str_min='0', str_sec='0',
...
-def calc_freq(index):
-    """ Determine the frequency for a time index
+def calc_res(index, use_mean=False):
+    """ Determine the resolution for a time index
 
     Parameters
     ----------
     index : array-like
         Datetime list, array, or Index
+    use_mean : bool
```
rstoneback, line 84: Would it be practical to support mode and median as well? Perhaps allow users to pass in their own favorite function or something.

Since the last review I've been a little uncomfortable with only a single averaging function since instrument outputs can all have all kinds of time jumps. We won't get all of them ahead of time but a user should have better knowledge of the specific data issues they face.

rstoneback (commented):
Thanks @aburrell. I didn't get a chance to dig into the common_index yet, per comments in the review, so that area may get some additional future feedback.

pysat/_constellation.py
```diff
@@ -33,17 +50,25 @@ class Constellation(object):
...
-    def __init__(self, const_module=None, instruments=None):
+    def __init__(self, const_module=None, instruments=None, index_res=None,
+                 common_index=True):
         """
         Constructs a Constellation given a list of instruments or the name of
         a file with a pre-defined constellation.
 
         Parameters
         ----------
         const_module : string
```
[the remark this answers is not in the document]
aburrell: Yes, that's intended, so this needs to be updated.

pysat/tests/test_constellation.py
```diff
@@ -104,27 +135,132 @@ def double_mlt(inst):
...
+            assert (inst['double_mlt'] == 2.0 * inst['mlt']).all()
+        return
+
+
+class TestConstellationFunc:
+    """Test the Constellation class attributes and methods."""
+    def setup(self):
+        """Create instruments and a constellation for each test
+        """
+        self.inst = list(constellations.testing.instruments)
```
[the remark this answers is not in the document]
aburrell: Could you put this as an issue?  It's out of scope here.

pysat/_constellation.py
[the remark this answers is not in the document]
aburrell: There is a flag in the `_empty` method that says it's not empty with partial data, but I decided to set it to be empty unless all Instruments have data.  Why don't you raise this as an issue and we can discuss it at the next meeting?

pysat/_constellation.py
```diff
@@ -136,6 +279,62 @@ def bounds(self, value=None):
...
+        # Determine which instrument variables share the same name
+        data_vars = dict()
+        for inst in self.instruments:
+            for dvar in inst.variables:
+                if dvar in data_vars.keys():
+                    data_vars[dvar] += 1
+                else:
+                    data_vars[dvar] = 1
+
+        # Distinguish shared names by Instrument platform, name, tag, inst_id
```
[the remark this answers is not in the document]
aburrell: Because it has whitespace and will be hard to use as a future data variable value.  I also hesitated to use `-` to separate.  Open to other suggestions, but keep in mind that in a file created after running the future `const_to_inst` these will be data variable names.

pysat/_constellation.py
```diff
@@ -136,6 +279,62 @@ def bounds(self, value=None):
...
+    def tomorrow(self):
+        """Returns UTC date for tomorrow, see pysat.Instrument for details."""
+        return self.today() + dt.timedelta(days=1)
+
+    def yesterday(self):
+        """Returns UTC date for yesterday, see pysat.Instrument for details."""
+        return self.today() - dt.timedelta(days=1)
+
+    @property
+    def variables(self):
```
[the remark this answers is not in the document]
aburrell: Well, it's really a discussion of the purpose and scope of this variable.  My thought process was:
1) This is a high level look at the data in the Constellation, and so should let people easily see the data.  Longer variable names means that less data variables can be easily displayed.
2) You need the instrument list index, not the platform/name/etc to go get the data, so the platform/name/etc is not as useful as it first appears.
3) It doesn't easily distinguish between overlapping and distinct data variables in the constellation.

That doesn't mean there are opposing reasons, but these are my reasons for not doing what you suggested.

pysat/_constellation.py
aburrell: See the above discussion for the `empty` flag.  Obviously, the name `empty` was used for compatibility with Instrument.

pysat/utils/time.py
aburrell: It's never practical to support the mode.  It wouldn't be hard to allow users to pass their own function in, but the question for this might be: "do we need this functionality?".  This should be raised as an issue, because it's well beyond the scope here.

pysat/_constellation.py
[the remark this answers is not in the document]
aburrell: Ok, having read more of your thoughts about the `empty` flag as they occurred in the review, I think what you want is a `partial` flag that would be `True` if you want `empty` to treat partial data as full and `False` for the current behaviour.  Actually, I think what you really want is a merged Instrument, but that's step 2. 👽  The framework for this exists, but again, it's a new thing and I think it should be a new issue.

pysat/tests/test_constellation.py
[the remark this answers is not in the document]
rstoneback: https://github.com/pysat/pysat/issues/798

pysat/_constellation.py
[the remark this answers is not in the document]
rstoneback: Yeah, I think the first part is correct, that I think that `empty` should have a different default for `all_inst`.

Not sure what you mean by merged instrument by interested to hear your thoughts at the next meeting.

pysat/_constellation.py
rstoneback: I'm with you on having a compatible `empty` but my built in expectation was/is different than how it currently operates.

rstoneback (approved):

aburrell:
Thanks for the thorough review @rstoneback!  I also like how most of it got pawned off into issues for small scale improvements 😸

Issue #541, opened 2020-09-04
Add method `set_inst_attr` to Constellation


aburrell:
**Is your feature request related to a problem? Please describe.**
The Constellation class has a method called `set_bounds` that sets the `bounds` attribute in each instrument.  This could be generalized to set other instrument attributes.

**Describe the solution you'd like**
Add a method `set_inst_attr` that takes the attribute desired value as input.

**Describe alternatives you've considered**
Not doing this.  However, I think it's only a good idea to leave things as is if `bounds` is the only attribute that should be the same across all instruments.

PR #841, opened 2021-07-09
Added `_set_inst_attr` to Constellation

aburrell:
# Description

Addresses #541 by adding a hidden method to set an instrument attribute to the same value across all instruments.  Also improved the Instrument docstrings.

## Type of change

# How Has This Been Tested?

Ran unit tests and added new/missing attributes to the attribute test.

**Test Configuration**:
* Version number: Python 3.9

# Checklist:

### DIFF
--- a/pysat/_constellation.py
+++ b/pysat/_constellation.py
@@ -71,11 +71,25 @@ class Constellation(object):
         data, or a tuple of NoneType objects. Users may provide as a tuple or
         tuple of lists (useful for bounds with gaps). The attribute is always
         stored as a tuple of lists for consistency.
+    date : dt.datetime or NoneType
+        Date and time for loaded data, None if no data is loaded
+    yr : int or NoneType
+        Year for loaded data, None if no data is loaded
+    doy : int or NoneType
+        Day of year for loaded data, None if no data is loaded
+    yesterday : dt.datetime
+        Date and time for yesterday in UT
+    today : dt.datetime
+        Date and time for the current day in UT
+    tomorrow : dt.datetime
+        Date and time for tomorrow in UT
     empty : bool
         Flag that indicates all Instruments do not contain data when True.
     empty_partial : bool
         Flag that indicates at least one Instrument in the Constellation does
         not have data when True.
+    variables : list
+        List of loaded data variables for all instruments.
 
     Raises
     ------
@@ -103,6 +117,9 @@ class Constellation(object):
         # Initalize the `instruments` attribute to be an empty list before
         # loading using each of the available input methods
         self.instruments = []
+        self.date = None
+        self.yr = None
+        self.doy = None
 
         # Add any registered Instruments that fulfill the provided platforms,
         # names, tags, and inst_ids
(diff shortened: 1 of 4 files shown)


pysat/_constellation.py
```diff
@@ -555,9 +586,16 @@ def load(self, *args, **kwargs):
...
         """
 
+        # Load the data for each instrument
         for instrument in self.instruments:
             instrument.load(*args, **kwargs)
 
+        # Set the year and doy attributes for the constellation and instruments
+        self.yr, self.doy = utils.time.getyrdoy(self.date)
+
+        self._set_inst_attr('yr', self.yr)
```
rstoneback, line 596: `yr` and `doy` are set on each Instrument as they are loaded. `self.date` maps back to `_index` which requires all instruments to have data, correct?

So if only some instruments have data then doesn't this lead to each instrument having a `yr` and `doy` of None?

pysat/_instrument.py
```diff
@@ -91,10 +92,23 @@ class Instrument(object):
...
+    clean_level
+    pad
+    orbit_info
+    inst_module
+    temporary_file_list
+    strict_time_flag
+    directory_format
+    file_format
+    bounds : tuple
+        Tuple of two datetime objects or filenames indicating bounds for loading
```
rstoneback, line 108:
```suggestion
        Tuple of datetime objects or filenames indicating bounds for loading
```
You mention the case of bounds with gaps, which requires more than two total objects. I can see specifying two as it is the simplest case, could help users out. At the same time, specifying two when in general it isn't could be confusing for a different class of users. Raising this for consideration rather than a required change.

pysat/_instrument.py
```diff
@@ -2062,6 +2094,10 @@ def custom_attach(self, function, at_pos='end', args=[], kwargs={}):
         return anything.
 
         """
+        if args is None:
```
rstoneback, line 2097: 👍

rstoneback (commented):
Thanks for the general standards upgrade as well as the new method.

rstoneback:
To go along with this PR and associate issue, I made an issue for a similar solution for Instrument methods https://github.com/pysat/pysat/issues/842

I scanned through the issue list and didn't see a pre-existing item.

pysat/_constellation.py
aburrell: `self.date` is `const.date` which maps back to the common index.  So I did this but wanted to see what you thought.  As is, if no data is loaded for an instrument, it will have `inst.date=None` and `inst.yr=YEAR; inst.doy=DOY` for the data loaded elsewhere.  Not sure we should be doing it, but thought it would be good to have things reletively consistent across all instruments.  Also totally OK doing away with this if it gives you worries about load/makes you grimice 🦇

rstoneback: I think changing the Instrument values should be avoided unless the method is specifically for changing those values. So I do have heartburn about that part. 

I like the addition of yr and doy to the constellation. Also makes sense to assign those directly in the constellation load.

I spent some time writing out a long thing about const.date but that isn't what this pull is about so I took it out instead. It may resurface as an issue.

pysat/_constellation.py
```diff
@@ -555,9 +586,16 @@ def load(self, *args, **kwargs):
...
+        self._set_inst_attr('doy', self.doy)
```
aburrell, line 597:
```suggestion
```

aburrell:
Ok @rstoneback, I think this PR is up to your specs now.

rstoneback:
Thanks @aburrell. Looks good. Was gonna push the approve button but figured I should wait until the tests pass, for standards sake and all that :)

rstoneback (approved):

Issue #887, opened 2021-08-19
BUG: xarray may leave a link to file open


jklenzing:
**Describe the bug**
When loading xarray data via `utils.load_netcdf`, a link to the file may be left open on some windows systems.  This creates issues  in the unit tests observed in pysat/pysatModels#89

**To Reproduce**
Look through logs associated with pysat/pysatModels#92

**Desktop (please complete the following information):**
 - OS: Windows

**Additional context**
Not noticing problems in similar xarray instruments for pysatNASA, but could be slight version differences.

jklenzing:
Following up on this, I've compared the pysatNASA and pysatModels log for CI tests run recently on Windows with python 3.8.  Both use the same versions of xarray (0.19.0), netCDF4 (1.5.7), and pysat (3.0.1).  Not sure why the ICON xarray instruments work and the model xarray instruments fail.  The pysatModels log does show a different version of the github actions runner (2.280.3 instead of 2.280.2).

Adding `out.close()` after the following load statements in the core package
https://github.com/pysat/pysat/blob/3d33e81de2a32da9f0deea2e8a5160aa017607cf/pysat/utils/_core.py#L404-L407
should fix the issue anywhere it occurs rather than fixing it after loading in other packages.

jklenzing:
Further testing with pysatNASA and the updated runner are working.

rstoneback:
Nice find!

On Aug 19, 2021, at 3:40 PM, Jeff Klenzing ***@***.******@***.***>> wrote:

Adding out.close() after the following load statements in the core package

[...]

PR #889, opened 2021-08-24
BUG: close files through xarray

jklenzing:
# Description

Addresses #887

On some windows systems, xarray leaves a link to the file open, which can create problems in the unit test suite.  This closes the file after loading the data.

## Type of change

# How Has This Been Tested?

via pytest on windows

# Checklist:

### DIFF
--- a/pysat/utils/_core.py
+++ b/pysat/utils/_core.py
@@ -410,6 +410,8 @@ def load_netcdf4(fnames=None, strict_meta=False, file_format=None,
         else:
             out = xr.open_mfdataset(fnames, combine='by_coords',
                                     decode_timedelta=decode_timedelta)
+        # Close any open links to file through xarray.
+        out.close()
         for key in out.variables.keys():
             # Copy the variable attributes from the data object to the metadata
             meta_dict = {}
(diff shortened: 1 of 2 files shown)


pysat/utils/_core.py
```diff
@@ -410,6 +410,8 @@ def load_netcdf4(fnames=None, strict_meta=False, file_format=None,
         else:
             out = xr.open_mfdataset(fnames, combine='by_coords',
                                     decode_timedelta=decode_timedelta)
+        # Close any open links to file through xarray.
+        out.close()
```
rstoneback, line 414: It seems a little weird that we can close the file so early and everything still works. So xarray must actually load everything from disk with the commands above. I wonder what the remaining open link to the file is for.

I would expect the close to be on line 435.

rstoneback (commented):
Thanks @jklenzing. I think I'd feel a bit more comfortable about future compatibility if the `.close` line was a bit further down in the code, after we are done accessing parameters.

pysat/utils/_core.py
jklenzing: The open commands should load everything from the file.  Moved as suggested in case it behaves differently on different systems.

aburrell (approved):

PR #893, opened 2021-09-01
STY: Reduce Code Duplication across projects for user libraries

jklenzing:
# Description

Makes `InstTestClass` more portable to streamline user implementation of the standard end-to-end instrument tests.
- Moves `setup_class` and `teardown_class` to `InstTestClass` so users do not have to copy these over with small changes.
- Adds `apply_marks_to_tests` method to class, which will autogenerate instrument lists and apply pytest marks appropriately.
- Updates the documentation for downstream developers accordingly.

Note that the current implementations in the library packages should still succeed with this version, since they duplicate the same code they already duplicated.  Changes to those codes should not be finalized until this version is released.

## Type of change

# How Has This Been Tested?

Local tests can be verified by running 
```
pytest -vs pysat/tests/test_instruments.py
```
Inheritance in an external package can be tested by running 
```
pytest -vs pysatNASA/tests/test_instruments.py
```
for the branch posted to pysat/pysatNASA#85.  The code there is an example of how the implementation is improved for downstream users.  It will not pass on Github Actions yet, so the branch must be checked out locally and run.

Successful run: all tests are run and succeed.

**Test Configuration**:
* Operating system: Mac OS X 10.15.7
* pytest 6.2.4

# Checklist:

Backwards-compliance checks
- [x] pysatCDAAC
- [x] pysatIncubator -- test_instruments working
- [x] pysatMadrigal
- [x] pysatMissions
- [x] pysatModels -- test_instruments working
- [x] pysatNASA
- [x] pysatSpaceWeather


pysat/tests/instrument_test_class.py
```diff
@@ -47,17 +49,93 @@ def initialize_test_inst_and_date(inst_dict):
...
+        # Define standard attributes to check.
+        self.module_attrs = ['platform', 'name', 'tags', 'inst_ids',
+                             'load', 'list_files', 'download']
+        self.inst_attrs = ['tag', 'inst_id', 'acknowledgements', 'references',
+                           'inst_module']
+        self.inst_callable = ['load', 'list_files', 'download', 'clean',
+                              'default']
+        self.attr_types = {'platform': str, 'name': str, 'tags': dict,
+                           'inst_ids': dict, 'tag': str, 'inst_id': str,
+                           'acknowledgements': str, 'references': str}
```
jklenzing, line 77: Reminding myself that moving these here will probably break things.  These could be moved back, or moved to a method-level setup.

jklenzing: These need to be defined outside the `setup_class` for backwards compatibility.  Comment added.

pysat/tests/instrument_test_class.py
```diff
@@ -47,18 +49,95 @@ def initialize_test_inst_and_date(inst_dict):
...
+                    getattr(self, method).pytestmark.append(mark)
+                elif 'download' in mark_names:
+                    mark = pytest.mark.parametrize("inst_dict",
+                                                   instruments['download'])
+                    getattr(self, method).pytestmark.append(mark)
+                elif 'no_download' in mark_names:
+                    mark = pytest.mark.parametrize("inst_dict",
+                                                   instruments['no_download'])
+                    getattr(self, method).pytestmark.append(mark)
+        return
```
jklenzing, line 139:
```suggestion
        return instruments
```
We could return the `instruments` object if needed so that it doesn't need to be run separately when needed for additional tests.  This would clean up the code, but may be more confusing when writing the docs.  Thoughts?

pysat/tests/test_instruments.py
```diff
@@ -34,70 +33,32 @@
...
-        pysat.params['data_dirs'] = self.saved_path
-        self.tempdir.cleanup()
-        del self.inst_loc, self.saved_path, self.tempdir
-        return
-
     # Custom package unit tests can be added here
 
-    # Custom Integration Tests added to all instruments.
+    # Custom Integration Tests added to all test instruments in core package.
     @pytest.mark.parametrize("inst_dict", [x for x in instruments['download']])
```
jklenzing, line 62:
```suggestion
    @pytest.mark.parametrize("inst_dict", InstTestClass.[x for x in instruments['download']])
```
Conversely, we could attach the instrument list to the test class directly and then invoke them in pytest like this.

jklenzing: Thinking about this from how I would explain the implementation to a new user, I think this is the best option, but open to feedback.

jklenzing:
@aburrell, this could go in as is, or we could simplify it slightly more via one of the two options above.  let me know if you have a preference.

pysat/tests/test_instruments.py
aburrell: I am fine either way, I keep going back and forth between liking one over the other.  I am curious as to whether you need to create a list from an existing list, though.

docs/dependency.rst
```diff
@@ -122,61 +196,23 @@ generally be unchanged.  Instruments are grouped in three lists:
...
-        # to point to their own subpackage location, e.g.,
-        # self.inst_loc = mypackage.instruments
-        self.inst_loc = pysat.instruments
-
-    def teardown_class(self):
-        """Runs once to clean up testing from this class."""
-        pysat.params['data_dirs'] = self.saved_path
-        self.tempdir.cleanup()
-        del self.inst_loc, self.saved_path, self.tempdir
+  @pytest.mark.parametrize("inst_dict", [x for x in instruments['download']])
```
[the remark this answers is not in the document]
jklenzing: That's a good catch.  Putting the list in directly here works.

pysat/tests/instrument_test_class.py
jklenzing: Using the return statement.

jklenzing:
Re-ran the backwards compatibility tests with all the packages listed above.  No issues in the instrument tests.

aburrell (approved):

PR #886, opened 2021-08-18
BUG/ENH: decode_timedelta for loading into xarray

jklenzing:
# Description

Addresses #823

Starting in `xarray` 0.16, loading from netcdf checks for time-like units and automatically converts the data into `dt.timedelta`s.  Adding the `decode_timedelta=False` kwarg to the load statements to maintain data format as default, but users may create instruments that pass this kwarg through if desired.

Updates the instrument template with default behavior for xarray and how to modify.

Note: The unit tests in this class will be rewritten as part of #60.  The current writeup is a workaround for the inability to write to netcdf4 directly from pysat.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update

# How Has This Been Tested?

New unit test captures behavior.

**Test Configuration**:
* Operating system: Mac OS X 10.15.7
* Version number: Python 3.8.2
* Any details about your local setup that are relevant

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``main``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] Add a note to ``CHANGELOG.md``, summarizing the changes

If this is a release PR, replace the first item of the above checklist with the release
checklist on the wiki: https://github.com/pysat/pysat/wiki/Checklist-for-Release

Issue #789, opened 2021-05-18
Remove support for `children` in Meta


rstoneback:
**Is your feature request related to a problem? Please describe.**
Support for children was added since multi-dimensional data originally required DataFrames within DataFrames. One variable label could have multiple variables stored within. pysat now supports xarray and this variable within variables issue should no longer exist.

**Describe the solution you'd like**
Add a DeprecationWarning.... and then remove support in the future.

**Describe alternatives you've considered**
Leave as is.

**Additional context**

jklenzing:
Deprecation in 3.1.0

jklenzing:
@rstoneback, are the 2D pandas objects being deprecated as well?  Or are we just changing how the meta is treated?

rstoneback:
You mean the DataFrames within DataFrames? Yes, deprecated. We shouldn’t be using that going forward.

jklenzing:
Proposal for next pysat meeting: We continue with meta children deprecation in 3.1.0, and push off other meta upgrades to 3.2.0 (when we remove the meta children).  This should make most of the rewrites simpler, since we won't have to deal with the higher order meta objects.

Issues potentially affected: #661, #767, #786, #900, #911
Pulls potentially affected: #896

Issue #60, opened 2018-07-07
netCDF4 and xarray


rstoneback:
xarray has built in support for netCDF4 files. pysat manages both data and metadata, so pysat's to_netCDF4 and from_netcdf4 need to be modified to include xarray as the data, while maintaining pysat's metadata support.

aburrell:
This is related to #78

jklenzing:
This is partially addressed in #441.  Still need the ability to write netcdf files from xarray.

jklenzing:
Bug found while working on #441 
```
import pysat 
testInst = pysat.Instrument(platform='pysat', 
                                             name='testing_xarray', 
                                             sat_id='100', 
                                             clean_level='clean')  
outfile = 'pysat_test_ncdf.nc'                                                                                                                                                         
testInst.to_netcdf4(outfile)    
loaded_inst, meta = pysat.utils.load_netcdf4(outfile, epoch_name='time', pandas_format=False)     
```
loaded_inst and testInst.data should be identical, but the 'time' coordinate is replaced by an 'Epoch' coordinate.  Trying to use the `epoch_name='time'` flag in `to_netcdf4` produces a netCDF4 error.
```
RuntimeError: NetCDF: String match to name in use
```

rstoneback:
> 
> ```
> RuntimeError: NetCDF: String match to name in use
> ```

to_netcdf4 writes the index information to file first using the label given in epoch_name.
If you give to_netcdf that the epoch name is 'time' and there is also a variable 'time' in the loaded array, then when the method tries to write the variable 'time' it will already exist.

rstoneback:
by default to_netcdf4 uses 'Epoch' for the time index label in files.

jklenzing:
> > ```
> > RuntimeError: NetCDF: String match to name in use
> > ```
> 
> to_netcdf4 writes the index information to file first using the label given in epoch_name.
> If you give to_netcdf that the epoch name is 'time' and there is also a variable 'time' in the loaded array, then when the method tries to write the variable 'time' it will already exist.

Got it.  So we should exclude the epoch_name from the variables when writing?

rstoneback:
> Got it. So we should exclude the epoch_name from the variables when writing?

I think that would solve it. We should either pop up a direct warning if we are removing the extra variable or raise an error and force the user to sort it out.

jklenzing:
[...]

PR #896, opened 2021-09-09
Add io utils

aburrell:
# Description

Pull request to ask for help and have a discussion on structure.

## Type of change

- Breaking change (fix or feature that would cause existing functionality
      to not work as expected)

# How Has This Been Tested?

Running `test_utils.py` unit tests (2DPandas and Xarray version of the netCDF tests currently fail).

**Test Configuration**:
* Version number: Python 3.9

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -1028,14 +1028,14 @@ class Meta(object):
             if lower_name == out_name.lower():
                 return out_name
 
-        # check if attribute present in higher order structures
+        # Verify that the attribute is present in higher order structures
         for key in self.keys_nD():
             for out_name in self[key].children.attrs():
                 if lower_name == out_name.lower():
                     return out_name
 
-        # nothing was found if still here
-        # pass name back, free to be whatever
+        # Nothing was found if still here. Pass `name` back without alteration,
+        # it is free to be whatever it is.
         return name
 
     def concat(self, other_meta, strict=False):
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -550,7 +550,7 @@ class Meta(object):
 
                 return meta_row
             else:
-                raise KeyError('Key not found in MetaData')
+                raise KeyError("Key '{:}' not found in MetaData".format(key))
         else:
             raise NotImplementedError("".join(["No way to handle MetaData key ",
                                                "{}; ".format(key.__repr__()),
@@ -1060,7 +1060,7 @@ class Meta(object):
 
         """
         # Cycle through the top-level variables
-        for var in self.attrs():
+        for var in self.keys():
             if isinstance(mapper, dict):
                 map_var = mapper[var]
             else:
@@ -1069,11 +1069,11 @@ class Meta(object):
             # Update the attribute name
             hold_meta = self[var].copy()
             self.drop(var)
-            self[mapper[var]] = hold_meta
+            self[map_var] = hold_meta
 
         # Determine if the attribute is present in higher order structures
         for ndkey in self.keys_nD():
-            for var in self[ndkey].children.attrs():
+            for var in self[ndkey].children.keys():
                 if isinstance(mapper, dict):
                     map_var = mapper[var]
                 else:
@@ -1081,8 +1081,8 @@ class Meta(object):
 
                 # Update the attribute name
                 hold_meta = self[ndkey].children[var].copy()
-                self[ndkey].children.drop(var)  # THIS DOESN'T WORK HERE
-                self[ndkey].children[mapper[var]] = hold_meta
+                self[ndkey].children.drop(var)
+                self[ndkey].children[map_var] = hold_meta
 
         return
 
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -1023,12 +1023,15 @@ class Meta(object):
         variable name.
 
         """
+        # Get a case-insensitive version of the name
         lower_name = name.lower()
+
+        # Determine if the input name is one of the top-level attributes
         for out_name in self.attrs():
             if lower_name == out_name.lower():
                 return out_name
 
-        # Verify that the attribute is present in higher order structures
+        # Determine if the attribute is present in higher order structures
         for key in self.keys_nD():
             for out_name in self[key].children.attrs():
                 if lower_name == out_name.lower():
@@ -1037,6 +1040,51 @@ class Meta(object):
         # Nothing was found if still here. Pass `name` back without alteration,
         # it is free to be whatever it is.
         return name
+
+    def rename(self, mapper):
+        """Returns preserved case name for case insensitive value of name.
+
+        Parameters
+        ----------
+        mapper : dict or func
+            Dictionary with old names as keys and new names as variables or
+            a function to apply to all names
+
+        Note
+        ----
+        Checks first within standard attributes. If not found there, checks
+        attributes for higher order data structures. If not found, returns
+        supplied name as it is available for use. Intended to be used to help
+        ensure that the same case is applied to all repetitions of a given
+        variable name.
+
+        """
+        # Cycle through the top-level variables
+        for var in self.attrs():
+            if isinstance(mapper, dict):
+                map_var = mapper[var]
+            else:
+                map_var = mapper(var)
+
+            # Update the attribute name
+            hold_meta = self[var].copy()
+            self.drop(var)
+            self[mapper[var]] = hold_meta
+
+        # Determine if the attribute is present in higher order structures
+        for ndkey in self.keys_nD():
+            for var in self[ndkey].children.attrs():
+                if isinstance(mapper, dict):
+                    map_var = mapper[var]
+                else:
+                    map_var = mapper(var)
+
+                # Update the attribute name
+                hold_meta = self[ndkey].children[var].copy()
+                self[ndkey].children.drop(var)  # THIS DOESN'T WORK HERE
+                self[ndkey].children[mapper[var]] = hold_meta
+
+        return
 
     def concat(self, other_meta, strict=False):
         """Concats two metadata objects together.
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -163,7 +163,7 @@ class Meta(object):
         # Set the labels
         self.labels = MetaLabels(metadata=self, **labels)
 
-        # init higher order (nD) data structure container, a dict
+        # Initialize higher order (nD) data structure container, a dict
         self._ho_data = {}
 
         # Use any user provided data to instantiate object with data
@@ -1191,6 +1191,34 @@ class Meta(object):
                                         'Instrument object.'))
                         raise RuntimeError(rerr)
         return
+
+    def add_epoch_metadata(self, epoch_name):
+        """Add epoch or time-index metadata if it is missing.
+
+        Parameters
+        ----------
+        epoch_name : str
+            Data key for time-index or epoch data
+
+        """
+        # Get existing meta data
+        if epoch_name in self:
+            new_dict = self[self.var_case_name(epoch_name)]
+        else:
+            new_dict = {}
+
+        # Update basic labels, if they are missing
+        epoch_label = 'Milliseconds since 1970-1-1 00:00:00'
+        basic_labels = [self.labels.units, self.labels.name, self.labels.desc,
+                        self.labels.notes]
+        for label in basic_labels:
+            if label not in new_dict or len(new_dict[label]) == 0:
+                new_dict[label] = epoch_label
+
+        # Update the meta data
+        self[self.var_case_name(epoch_name)] = new_dict
+
+        return  
 
     @classmethod
     def from_csv(cls, filename=None, col_names=None, sep=None, **kwargs):
--- a/pysat/tests/test_utils.py
+++ b/pysat/tests/test_utils.py
@@ -487,7 +487,7 @@ class TestLoadNetCDF4(object):
         return
 
 
-class TestLoadNetCDF4XArray(object):
+class TestLoadNetCDF4XArray(TestLoadNetCDF4):
     """Unit tests for `load_netcdf4` using xarray data.
 
     Note
--- a/pysat/tests/test_utils.py
+++ b/pysat/tests/test_utils.py
@@ -385,8 +385,7 @@ class TestLoadNetCDF4(object):
         self.testInst.load(date=self.stime)
 
         # Modify data and metadata names in data
-        self.testInst.meta.data = self.testInst.meta.data.rename(str.upper,
-                                                                 axis='index')
+        self.testInst.meta.rename(str.upper)
         if self.testInst.pandas_format:
             self.testInst.data = self.testInst.data.rename(str.upper,
                                                            axis='columns')
--- a/pysat/utils/_core.py
+++ b/pysat/utils/_core.py
@@ -151,6 +151,10 @@ def load_netcdf4(fnames=None, strict_meta=False, file_format=None,
                          'fill_val': ('fill', np.float64)}):
     """Load netCDF-3/4 file produced by pysat.
 
+    .. deprecated:: 3.2.0
+       Function moved to `pysat.utils.io`, this wrapper will be removed in
+       the 3.2.0+ release.
+
     Parameters
     ----------
     fnames : str, array_like, or NoneType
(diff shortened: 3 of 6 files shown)


pysat/utils/_core.py
```diff
@@ -194,242 +199,21 @@ def load_netcdf4(fnames=None, strict_meta=False, file_format=None,
...
-        # Remove attributes from the data object
-        out.attrs = {}
+    out, meta = pysat.utils.io.load_netcdf4(fnames, strict_meta=strict_meta,
+                                            file_format=file_format,
+                                            epoch_name=epoch_name,
+                                            pandas_format=pandas_format,
+                                            labels=labels)
 
         # Close any open links to file through xarray.
         out.close()
```
jklenzing, line 219:
```suggestion
    out, meta = pysat.utils.io.load_netcdf4(fnames, strict_meta=strict_meta,
                                            file_format=file_format,
                                            epoch_name=epoch_name,
                                            pandas_format=pandas_format,
                                            labels=labels)

    # Close any open links to file through xarray.
    out.close()
```

pysat/utils/io.py
```diff
@@ -0,0 +1,1029 @@
...
+    # Initialize local variables
+    saved_meta = None
+    running_idx = 0
+    running_store = []
+    two_d_keys = []
+    two_d_dims = []
+    meta = pysat.Meta(labels=labels)
+
+    # Load data by type
+    if pandas_format:
```
jklenzing, line 326: To reduce complexity, we could add two hidden functions: `_load_netcdf_pandas` and `_load_netcdf_xarray` and then just choose which to run here.  The nesting in the pandas section is already getting fairly deep, and this would buy back some space as well as help when debugging for specific files.
```suggestion
    if pandas_format:
        out, meta = _load_netcdf_pandas(fname, format, etc)
    else:
        out, meta = _load_netcdf_xarray(fname, format, etc)
```

jklenzing (commented):
I like the reorganization idea.  A few lines were missed in the move.

pysat/utils/io.py
jklenzing: Lines below would need to be moved into the new functions

aburrell: I am ok doing this.

pysat/tests/test_utils.py
```diff
@@ -486,7 +485,7 @@ def test_netcdf_attribute_override(self):
         return
 
 
-class TestLoadNetCDF4XArray(object):
+class TestLoadNetCDF4XArray(TestLoadNetCDF4):
```
jklenzing, line 488: File output for xarray needs to be fixed before this can use inheritance.  See #60

pysat/utils/io.py
```diff
@@ -0,0 +1,1033 @@
...
+        for item in running_store:
+            out.append(pds.DataFrame.from_records(item, index=epoch_name))
+        out = pds.concat(out, axis=0)
+    else:
+        # The data is in xarray format, load differently for single or
+        # multiple files
+        if len(fnames) == 1:
+            out = xr.open_dataset(fnames[0])
+        else:
+            out = xr.open_mfdataset(fnames, combine='by_coords')
```
jklenzing, line 526:
```suggestion
        if len(fnames) == 1:
            out = xr.open_dataset(fnames[0], decode_timedelta=decode_timedelta)
        else:
            out = xr.open_mfdataset(fnames, combine='by_coords',
                                    decode_timedelta=decode_timedelta)
```

jklenzing: The bugfix here got dropped in the copy, but the tests to check this did get merged in, hence the errors for `test_utils.py::TestLoadNetCDF4XArray::test_read_netcdf4_with_time_meta_labels`

jklenzing: The kwarg `decode_time_delta` needs to be added to the functions and documentation.  I think this branch may predate #886.  See changes there for example

pysat/utils/io.py
```diff
@@ -0,0 +1,1033 @@
...
+                    # netCDF, to loaded data dictionary
+                    loaded_vars[obj_key] = loop_list
+                    del loop_list
+
+                # Prepare dataframe index for this netcdf file
+                time_var = loaded_vars.pop(epoch_name)
+
+                # Convert from GPS seconds to seconds used in pandas (unix time,
+                # no leap)
+                # time_var = convert_gps_to_unix_seconds(time_var)
```
aburrell, line 502: It looks like this is no longer being done?  @rstoneback @jklenzing

jklenzing: I'm fine dropping this.  @rstoneback?

pysat/utils/io.py
aburrell: Done!

pysat/tests/test_utils.py
aburrell: This is going to be performed in this branch, so will be kept in here.  One of the things I'm asking for help with.

pysat/utils/io.py
aburrell: Fixed!

pysat/utils/io.py
```diff
@@ -0,0 +1,1159 @@
...
+    out = []
+    for item in running_store:
+        out.append(pds.DataFrame.from_records(item, index=epoch_name))
+    data = pds.concat(out, axis=0)
+
+    return data, meta
+
+
+def load_netcdf_xarray(fnames, strict_meta=False, file_format='NETCDF4',
+                       epoch_name='Epoch', decode_deltatime=False,
```
jklenzing, line 594:
```suggestion
                       epoch_name='Epoch', decode_timedelta=False,
```
Here and elsewhere

aburrell:
Ok @jklenzing that fixed two of the unit tests.  No idea why the 2D Pandas stuff isn't working though 😿  🐛

jklenzing:
Was just looking into 🔲 🐼  tests.  They're working under develop, probably just need to compare that the function move is handling things correctly.

aburrell:
> Was just looking into 🔲 🐼 tests. They're working under develop, probably just need to compare that the function move is handling things correctly.

Yeah, I did a line-by-line check with develop, created files from both routines, and could not find where the new routine is going wrong.

jklenzing:
The str characters under the loaded_inst are being imported as 't', not 'test'.  Not sure whether this is an issue on the read or write side

jklenzing:
Verified it's on the write side.  Repeated steps for `test_write_and_read_netcdf4_w_kwargs` in console, but loaded data directly into netcdf without using pysat.  `testInst.data['profiles'][0]['dummy_str']` is being written as "t", not "test" for each value.

pysat/utils/io.py
```diff
@@ -0,0 +1,1159 @@
...
+                                    pysat.logger.info(
+                                        ''.join(('Unable to find MetaData for ',
+                                                 lkey)))
+
+                                # Attach data.  It may be slow to repeatedly
+                                # call the store method as well astype method
+                                # below collect data into a numpy array, then
+                                # write the full array in one go
+                                temp_cdf_data = np.zeros(
+                                    (num, dims[0]), dtype=coltype)
```
jklenzing, line 1028:
```suggestion
                                temp_cdf_data = np.zeros(
                                    (num, dims[0])).astype(coltype)
```
Not sure why, but using dtype only allows one-character strings to be put in the array.  This is the source of 5 out of 6 of the pandas 25 bugs.  🐼 🐛

jklenzing: If coltype=='str', we could specify `dtype='U32'` to maintain behavior but eliminate `astype`.

But 2D pandas is going to be deprecated soon, so optimization of these codes is probably not high priority.

jklenzing:
After the dtype issue is sorted, the remaining 2D pandas error is in `test_basic_write_and_read_netcdf4_mixed_case_meta_format`:
```
 'alt_profiles' not in Index(['ALTITUDE', 'ALT_PROFILES', ...
```
This works for the 1D case, so whatever is supposed to maintain the case on 2D variables is not working in the writing routine.

aburrell:
I just tried implementing that and all the same tests fail locally.  I think it is a worthwhile change to make.  If you've done it locally can you push it?

I think the remaining test that's failing is failing because of the new `Meta.rename` method.  Hopefully I can sort that out.

jklenzing:
The failure in `TestLoadNetCDF4XArray::test_netcdf_attribute_override` is caused by the test instruments improperly storing metadata as children objects for xarray.  This should be fixed on the test instrument side.

jklenzing:
@aburrell, I can work on the metadata changes here or make a pull into this branch.  Any preferences?

aburrell:
@jklenzing go ahead and work on it here!  I have a small change I will push first.

aburrell:
All yours @jklenzing 🐈

jklenzing:
> @jklenzing go ahead and work on it here! I have a small change I will push first.

OK.  Changes will primarily be in the pysat/instruments folder.

jklenzing:
@aburrell, I've updated the meta accordingly and marked lines that need to be removed or changed when #789 is implemented.  The errors I see now are coming from xarray, so at least we've moved it outside of the `is_fill` method.

jklenzing:
Getting closer!  The xarray tests are now failing because some of the tests will only work for pandas:
https://github.com/pysat/pysat/blob/bf30e010b387512de6c58553b9a522fb386ad09d/pysat/tests/test_utils.py#L365-L368

`xarray.Dataset` objects do not have columns, so these need to be generalized.

jklenzing:
tentative fix locally.  will finish later

jklenzing:
@aburrell, it's down to 4 errors.  I've removed the one extra test under `TestLoadNetCDF4XArray` that was designed to compensate for the lack of pysat write statements.

Also, I've pulled out the sorting statements since we don't really care about that under `self.eval_loaded_data` and generalizing them to work with xarray is a lot of extra work.

jklenzing:
Down to 3 errors, which are all instances of the same test: `test_basic_write_and_read_netcdf4_mixed_case_meta_format`.

Upper case variables names are not being written to file.

aburrell:
And the errors are consistent across data types!  This is great work @jklenzing 😸 I am tackling it now.

pysat/tests/test_utils.py
```diff
@@ -321,7 +321,13 @@ def teardown(self):
     def eval_loaded_data(self):
         """Evaluate loaded test data."""
         # Test that the written and loaded data matches the initial data
-        for dkey in self.testInst.data.columns:
+        if self.testInst.pandas_format:
```
rstoneback, line 324:
```suggestion
       for dkey in self.testInst.variables:
```
should work for both pandas and xarray.

pysat/utils/io.py
```diff
@@ -0,0 +1,1159 @@
...
+    # Ensure the metadata is set and updated to netCDF4 standards
+    add_netcdf4_standards_to_meta(inst, epoch_name)
+
+    # Base_instrument used to define the standard attributes attached
+    # to the instrument object. Any additional attributes added
+    # to the main input Instrument will be written to the netCDF4
+    if base_instrument is None:
+        base_attrb = dir(pysat.Instrument())
+    else:
+        warnings.warn("".join(["`base_instrument` has been deprecated and will",
```
rstoneback, line 768: Why is base_instrument being deprecated?

pysat/utils/io.py
rstoneback: I think this was already dropped while I was 'away.' Looked into the history and the commented line was never uncommented here. I think this section likely originated from the COSMIC project and this is a leftover. Thanks for the cleanup.

rstoneback (commented):
Thanks @aburrell and @jklenzing. This is a partial review pass.

pysat/tests/test_utils.py
aburrell: The `loaded_inst` object varies, so we still need the if/else statement.

aburrell: I tested moving the `keys` eval outside the if/else loop and it failed.

pysat/utils/io.py
aburrell: I didn't see why it was useful.  This is one of the reasons this is a draft PR.

aburrell:
Thanks for your help @jklenzing!  I got all the tests passing locally now.  Keeping this as a draft until we:
- [ ] reorg the unit tests appropriately
- [ ] add unit tests for deprecation warnings
- [ ] add unit tests for new functions and methods
- [ ] update the documentation

If you want to take on any of these tasks, edit this comment and put your name next to the item.

pysat/utils/io.py
```diff
@@ -0,0 +1,1156 @@
...
+    running_store = []
+    two_d_keys = []
+    two_d_dims = []
+    meta = pysat.Meta(labels=labels)
+
+    # Load data for each file
+    for fname in fnames:
+        with netCDF4.Dataset(fname, mode='r', format=file_format) as data:
+            # Build a dictionary with all global ncattrs and add those
+            # attributes to a pysat meta object
```
rstoneback, line 404:
```suggestion
            # Build a dictionary with all global ncattrs and add those
            # attributes to a pysat meta object. Any custom Meta attributes 
            # are automatically transferred to the holding pysat.Instrument 
            # object by pysat.
```
I went down a path here on Meta attributes and ended up verifying a design choice that I'd forgotten.

rstoneback (commented):
Thanks again @aburrell!

jklenzing:
> Thanks for your help @jklenzing! I got all the tests passing locally now. Keeping this as a draft until we:
> 
> * [ ]  reorg the unit tests appropriately
> * [ ]  add unit tests for deprecation warnings
> * [ ]  add unit tests for new functions and methods
> * [ ]  update the documentation
> 
> If you want to take on any of these tasks, edit this comment and put your name next to the item.

I'd suggest adding on a reorg for the io functions as well.  The `to_netcdf` function should be broken into `_to_netcdf_pandas` and `_to_netcdf_xarray`, similar to how it's handled for the load functions.  I lean toward making the specialized functions hidden to simplify things for end users, but open to discussion.

rstoneback:
> 
> I'd suggest adding on a reorg for the io functions as well. The `to_netcdf` function should be broken into `_to_netcdf_pandas` and `_to_netcdf_xarray`, similar to how it's handled for the load functions. I lean toward making the specialized functions hidden to simplify things for end users, but open to discussion.

I'm not so sure. For the two loads the methods return different data types. When writing files though the same output is intended. I also think the to_netcdf code should, as long as possible, stay a single path so that the resulting netcdfs are as close as possible to each other, independent of pandas or xarray. Breaking the methods up doesn't prevent that but it does cost more to get the same level of functionality.

aburrell:
Can we put this high on the agenda?  I don't want the various parts of this that need to be discussed to get too lost in the comment threads.

rstoneback:
> 
> If you want to take on any of these tasks, edit this comment and put your name next to the item.

Happy to contribute. I was thinking of getting our discussions sorted out first before signing up. 🐵

rstoneback:
> Can we put this high on the agenda? I don't want the various parts of this that need to be discussed to get too lost in the comment threads.

If you mean the agenda for the pysat meeting of course. I'll update it now.

rstoneback:
Awesome. Thanks @aburrell for kicking off the file writing re-org. It was not only needed but I think it'll make the `_instrument.py` file feel lighter, plus it forces us to re-evaluate what pysat does at the interface with files. Though this comment is for a later pull, I'm excited about getting the higher-order support out of the pandas file area. We'll just get to delete all those lines and the code will be easier for everyone to follow.

pysat/utils/io.py
jklenzing: TODO: add issue for new usage of base_instrument

pysat/tests/test_utils_io.py
```diff
@@ -0,0 +1,381 @@
...
+        if self.testInst.pandas_format:
+            self.testInst.data = self.testInst.data.rename(str.upper,
+                                                           axis='columns')
+        else:
+            map_keys = {dkey: dkey.upper()
+                        for dkey in self.testInst.data.data_vars.keys()}
+            self.testInst.data = self.testInst.data.rename(map_keys)
+
+        # Meta case is preserved and has not been altered
+        self.testInst.to_netcdf4(outfile, preserve_meta_case=True)
```
jklenzing, line 129: Should we switch to the new notation so we don't pile up deprecation warnings while running the tests?

aburrell: Yes, these are the tests of the future!  Let @JonathonMSmith handle the tests of the past 🦋

pysat/_meta.py
```diff
@@ -1021,21 +1021,99 @@ def attr_case_name(self, name):
...
+        Note
+        ----
+        Checks first within standard attributes. If not found there, checks
+        attributes for higher order data structures. If not found, returns
+        supplied name as it is available for use. Intended to be used to help
+        ensure that the same case is applied to all repetitions of a given
+        variable name.
+
+        """
+        def get_mapped_value(value, mapper):
```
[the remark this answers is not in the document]
aburrell: Ok, so I went to go do this and ran into the problem that this method doesn't use `self` at all.  So, I could put it as a hidden `_meta` method, keep it here, or I could put it somewhere in `utils`.  Not sure where it would go besides `_core`.  But `utils` may be the right place if we want to eventually add a `rename` method to `Instrument`.

pysat/_meta.py
```diff
@@ -1021,21 +1021,99 @@ def attr_case_name(self, name):
...
+            value : str
+                MetaData variable name to be adjusted
+            mapper : dict or function
+                Dictionary with old names as keys and new names as variables or
+                a function to apply to all names
+
+            Returns
+            -------
+            mapped_val : str or NoneType
+                Adjusted MetaData variablle name or NoneType if input value
```
rstoneback, line 1074:
```suggestion
                Adjusted MetaData variable name or NoneType if input value
```

pysat/_meta.py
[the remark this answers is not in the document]
rstoneback: Good point.

> Ok, so I went to go do this and ran into the problem that this method doesn't use `self` at all. 

Ah, right...

> So, I could put it as a hidden `_meta` method, keep it here, or I could put it somewhere in `utils`. 

Given that technical point I think where it is is fine then.

> if we want to eventually add a `rename` method to `Instrument`.

We got one :)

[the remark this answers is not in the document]
aburrell: > We got one :)

Ok, if we ever want to *upgrade* Instrument.rename. 😜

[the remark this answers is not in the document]
rstoneback: Made an issue to track that there is now technically two code paths in the repo for changing a metadata variable name. pysat.Instrument has its own code for renaming metavariables that should be updated to use meta.rename. #900

jklenzing:
Latest errors are a result of my updates.  Will put in a fix, and pull in latest develop changes to assist with #901.

jklenzing:
Fixed meta tests locally.  Some of the deprecated functions were used in `test_meta`.  I've updated the refs, and included the tempfile usage adopted elsewhere to keep pysat tests from altering a user's config.  The `remove_files` function in the tests is now gone.

There's still an issue in the `scale_units` tests. Verified in ipython that none of the tests for ` pysat.utils.scale_units(key, 'm/s')` for the new parameters. From the output, it looks like the `replace_str` routines aren't working as expected, since the failure is at this line:
```
>       unit_scale = scales[out_key] / scales[in_key]
E       KeyError: 'm s$^{-1}$'
```

pysat/utils/_core.py
```diff
@@ -92,15 +101,15 @@ def scale_units(out_unit, in_unit):
...
+    if out_key in ['m', 'm/s', 'm-3'] or in_key in ['m', 'm/s', 'm-3']:
         if in_key != out_key:
             raise ValueError('Cannot scale {:s} and {:s}'.format(out_unit,
                                                                  in_unit))
         # Recast units as keys for the scales dictionary and ensure that
         # the format is consistent
         rkey = ''
         for rr in replace_str.keys():
-            if out_key.find(rr):
+            if out_key.find(rr) >= 0:
```
jklenzing, line 112: Found the bug.  The previous line was always True, so whatever was last checked was selected.  This resulted in the following lines trying to replace velocity with volume.

jklenzing:
Note: pytest currently working, but coveralls upload is failing

jklenzing:
Coveralls is down for maintenance.  https://status.coveralls.io/

Issue #940, opened 2021-10-25
Easier to expand metalabels


rstoneback:
**Is your feature request related to a problem? Please describe.**
Suppose a user starts with a regular instrument and then wants to make a SPDF compatible file. Is there an easy way to update the meta labels so the user can fill in values for non-default pysat meta labels after an Instrument is instantiated? 

I'm in recording thoughts mode. Intension is to follow up with code review but other deadlines are closer.

**Describe the solution you'd like**
There is already support for accepting labels from another meta instance. I think we can break up that pair of functions, cannibalize some of the code, and make a function that accepts a labels instance and updates the labels for meta, which may include new labels. We need to keep the existing functionality at the meta class level as well, currently used by the loading process.

**Describe alternatives you've considered**
There is currently some built in expansion of meta as files are written, focused in areas of netcdf standards. We could exclusively expand within the `to_netcdf` realm but that leaves out some of the expanded generality with the recent `io` expansion.

**Additional context**
SPDF may be optimized for the file level, not as much at the human level. User focused to make it easy to go from one to the other in a systematic and general way. Always some balance between enabling features and making use more complicated. Things the user shouldn't have to worry about, like variable type and what not, should stay automatic.

PR #961, opened 2021-11-15
ENH: updated meta assignment

aburrell:
# Description

Updated the meta data assignment by checking type against expected label type, updating MetaLabels if needed, and *not* allowing array inputs.  Possibly addresses #940 and was meant to address https://github.com/pysat/pysatNASA/issues/99.

## Type of change

# How Has This Been Tested?

```
import pysat
test_inst = pysat.Instrument('pysat', 'testing')
test_inst.load(2009, 1)
test_inst.meta['dummy1'] = {'new meta': 'test string'}
test_inst.meta['dummy1'] = {'new meta': 'test string'}
test_inst.meta['dummy1'] = {'new meta': b'test bytes'}
test_inst.meta['dummy2'] = {'new_bytes': b'test bytes'}
test_inst.meta.labels.label_type

{'units': str,
 'name': str,
 'notes': str,
 'desc': str,
 'min_val': numpy.float64,
 'max_val': numpy.float64,
 'fill_val': numpy.float64,
 'new_meta': str,
 'new_bytes': str}
```

Now let's raise an error!
```
import numpy as np

test_inst.meta['dummy1'] = {'new_array': np.array([0, 1.0])}

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-05d199d2ff20> in <module>
----> 1 test_inst.meta['dummy1'] = {'new_array': np.array([0, 1.0])}

~/Programs/Git/pysat/pysat/_meta.py in __setitem__(self, data_vars, input_dat)
    418                             # Update the MetaLabels object and the existing
    419                             # metadata to ensure all data have all labels
--> 420                             self.labels.update(iattr, ikey, itype)
    421                             self._label_setter(ikey, ikey, type(to_be_set))
    422 

~/Programs/Git/pysat/pysat/_meta.py in update(self, lattr, lname, ltype)
   1788         else:
   1789             # This is an invalid meta data type, raise a TypeError
-> 1790             raise TypeError(''.join(['iterable types like ', repr(ltype),
   1791                                      ' (set for ', repr(lattr),
   1792                                      ') are not allowed']))

TypeError: iterable types like <class 'numpy.ndarray'> (set for 'new_array') are not allowed
```
**Test Configuration**:
* Version number: Python 3.8

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,11 +31,13 @@ This project adheres to [Semantic Versioning](https://semver.org/).
    * Added the ability to only download new files if remote file listing
      capabilities are not available for the Instrument.
    * Added kwargs for epoch units and origin in `pysat.utils.io.load_netCDF`.
-   * Vectorized `Meta.var_case_name` and `Meta.attr_case_name` to support 
+   * Vectorized `Meta.var_case_name` and `Meta.attr_case_name` to support
      list of str as input as well as str.
    * Added a time function to calculate decimal year from datetime.
    * Allow `Instrument.rename` to take a fuction or mapping dict as input,
      after adapting routine to use `Meta.rename`
+   * Added an update method and type evaluation method to MetaLabels.
+   * Allowed MetaLabels to be expanded through setting new Meta data values.
 * Deprecations
    * Removed `freq` as a standard kwarg for `pysat.Instruments.download`
    * Deprecated `_test_download_travis` as a standard attribute for
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -400,12 +400,36 @@ class Meta(object):
                                    and self.labels.label_type[iattr] == str):
                                     to_be_set = to_be_set.decode("utf-8")
                                 else:
-                                    warnings.warn(''.join((
-                                        'Metadata does not have expected type ',
-                                        repr(self.labels.label_type[iattr]),
-                                        '. Dropping input for  ', var,
-                                        ' with key ', ikey)))
-                                    good_set = False
+                                    # This type is incorrect, try casting it
+                                    wmsg = ''.join(['Metadata does not have ',
+                                                    'expected type ',
+                                                    repr(self.labels.label_type[
+                                                        iattr])])
+                                    try:
+                                        if hasattr(to_be_set, '__iter__'):
+                                            if self.labels.label_type[
+                                                    iattr] == str:
+                                                to_be_set = '\n\n'.join(
+                                                    [str(tval) for tval in
+                                                     to_be_set])
+                                            else:
+                                                raise TypeError("can't recast")
+                                        else:
+                                            to_be_set = self.labels.label_type[
+                                                iattr](to_be_set)
+
+                                        # Inform user data was recast
+                                        pysat.logger.warning(''.join((
+                                            wmsg, '. Recasting input for ',
+                                            repr(var), ' with key ',
+                                            repr(ikey))))
+                                    except (TypeError, ValueError):
+                                        # Warn user data was dropped
+                                        warnings.warn(''.join((
+                                            wmsg, '. Dropping input for ',
+                                            repr(var), ' with key ',
+                                            repr(ikey))))
+                                        good_set = False
                         else:
                             # Extend the meta labels. Ensure the attribute
                             # name has no spaces and that bytes are used instead
@@ -450,7 +474,7 @@ class Meta(object):
         elif isinstance(input_data, Meta):
             # Dealing with a higher order data set.
             # data_vars is only a single name here (by choice for support)
-            if (data_vars in self._ho_data) and (input_data.empty):
+            if (data_vars in self._ho_data) and input_data.empty:
                 # No actual metadata provided and there is already some
                 # higher order metadata in self.
                 return
@@ -964,13 +988,22 @@ class Meta(object):
 
         """
         # Update labels in metadata
-        for key in other_meta.labels.label_type:
+        for key in other_meta.labels.label_type.keys():
             new_name = getattr(other_meta.labels, key)
-            old_name = getattr(self.labels, key)
-            if old_name != new_name:
-                self._label_setter(new_name, old_name,
-                                   other_meta.labels.label_type[key],
-                                   use_names_default=True)
+
+            if hasattr(self.labels, key):
+                # This label exists, but the name may need updating
+                old_name = getattr(self.labels, key)
+                if old_name != new_name:
+                    self._label_setter(new_name, old_name,
+                                       other_meta.labels.label_type[key],
+                                       use_names_default=True)
+            else:
+                # This label doesn't exist, add it.
+                self.labels.update(key, new_name,
+                                   other_meta.labels.label_type[key])
+                self._label_setter(new_name, new_name,
+                                   other_meta.labels.label_type[key])
 
         self.labels = other_meta.labels
 
--- a/pysat/_meta.py
+++ b/pysat/_meta.py
@@ -386,26 +386,37 @@ class Meta(object):
                 if ikey not in ['children', 'meta']:
                     for i, var in enumerate(data_vars):
                         to_be_set = input_data[ikey][i]
-                        if hasattr(to_be_set, '__iter__') \
-                           and not isinstance(to_be_set, str):
-                            # We have some list-like object that can only
-                            # store a single element
-                            if len(to_be_set) == 0:
-                                # Empty list, ensure there is something to set
-                                to_be_set = ['']
-                            if isinstance(to_be_set[0], str) \
-                                    or isinstance(to_be_set, bytes):
-                                if isinstance(to_be_set, bytes):
-                                    to_be_set = to_be_set.decode("utf-8")
+                        good_set = True
 
-                                self._data.loc[var, ikey] = '\n\n'.join(
-                                    to_be_set)
-                            else:
-                                warnings.warn(' '.join(('Array elements are',
-                                                        'not allowed in meta.',
-                                                        'Dropping input for',
-                                                        var, 'with key', ikey)))
+                        # See if this meta data key has already been defined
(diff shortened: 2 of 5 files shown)


aburrell:
@rstoneback can you verify this does/does not fully address your issue that was linked?

rstoneback:
> @rstoneback can you verify this does/does not fully address your issue that was linked?

Yep. I will take a look today.

CHANGELOG.md
```diff
@@ -31,11 +31,13 @@ This project adheres to [Semantic Versioning](https://semver.org/).
...
      capabilities are not available for the Instrument.
    * Added kwargs for epoch units and origin in `pysat.utils.io.load_netCDF`.
-   * Vectorized `Meta.var_case_name` and `Meta.attr_case_name` to support 
+   * Vectorized `Meta.var_case_name` and `Meta.attr_case_name` to support
      list of str as input as well as str.
    * Added a time function to calculate decimal year from datetime.
    * Allow `Instrument.rename` to take a fuction or mapping dict as input,
      after adapting routine to use `Meta.rename`
+   * Added an update method and type evaluation method to MetaLabels.
+   * Allowed MetaLabels to be expanded through setting new Meta data values.
```
rstoneback, line 40:
```
inst.meta['test'] = {'units': 'yeah', 'new_label': 'old_info'}
```
works nicely!

Formatting is bad below, I also removed non-relevant variables.
```
In [6]: inst.meta
Out[6]: 
pysat.Meta(metadata=                 units            long_name                                              notes                                 desc value_min value_max fill new_label
        
test              yeah                 test                                                                                               NaN       NaN  NaN  old_info
```

pysat/_meta.py
```diff
@@ -386,26 +386,66 @@ def __setitem__(self, data_vars, input_dat):
...
+                                                    [str(tval) for tval in
+                                                     to_be_set])
+                                            else:
+                                                raise TypeError("can't recast")
+                                        else:
+                                            to_be_set = self.labels.label_type[
+                                                iattr](to_be_set)
+
+                                        # Inform user data was recast
+                                        pysat.logger.warning(''.join((
```
rstoneback, line 422: I tried this branch out on an ICON file. It leads to a whole long list of warnings.

```
In [8]: ivm = pysat.Instrument('icon', 'ivm', inst_id='a', update_files=True)

In [10]: ivm.load(2020, 1)
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'Epoch' with key 'FillVal'
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'ICON_L27_Altitude' with key 'FillVal'
pysat WARNING: Metadata does not have expected type <class 'numpy.float64'>. Recasting input for 'ICON_L27_Fractional_Ion_Density_H' with key 'Valid_Min'
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'ICON_L27_Fractional_Ion_Density_H' with key 'ValidMax'
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'ICON_L27_Fractional_Ion_Density_H' with key 'ValidMin'
pysat WARNING: Metadata does not have expected type <class 'numpy.float64'>. Recasting input for 'ICON_L27_Fractional_Ion_Density_H' with key 'Valid_Max'
```

Only showing a small fraction since it is way too much. Two contrasting requests:
 - From the error I don't know what type the data loaded as. I think adding that info may be helpful.
 - There are too many messages :(

rstoneback: Also tried it out on a MIGHTI file as a check to make sure it isn't just my files that produce warnings. It started to produce warnings but ran into an issue.

```
In [16]: mighti = pysat.Instrument('icon', 'mighti', inst_id='vector', tag='vector_wind_green')

In [19]: mighti.load(2020, 1)
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'Epoch' with key 'FillVal'
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'Epoch' with key 'ValidMin'
pysat WARNING: Metadata does not have expected type <class 'float'>. Recasting input for 'Epoch' with key 'ValidMax'
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-9f1312956ee0> in <module>
----> 1 mighti.load(2020, 1)

~/Code/pysat/pysat/_meta.py in __setitem__(self, data_vars, input_dat)
    442                             # Update the MetaLabels object and the existing
    443                             # metadata to ensure all data have all labels
--> 444                             self.labels.update(iattr, ikey, itype)
    445                             self._label_setter(ikey, ikey, type(to_be_set))
    446 

~/Code/pysat/pysat/_meta.py in update(self, lattr, lname, ltype)
   1821         else:
   1822             # This is an invalid meta data type, raise a TypeError
-> 1823             raise TypeError(''.join(['iterable types like ', repr(ltype),
   1824                                      ' (set for ', repr(lattr),
   1825                                      ') are not allowed']))

TypeError: iterable types like <class 'numpy.ndarray'> (set for 'Valid_Range') are not allowed

```

rstoneback (commented):
Thanks @aburrell! Yep, the code does support the type of feature described in #940. I don't think I originally considered the auto new label support (at high Meta level) you added but that is a nice, user-friendly touch! I tried out the other metadata changes on an ICON IVM and MIGHTI file and I think we need to turn down the sensitivity on when the warnings are generated. More below.

pysat/_meta.py
aburrell: So, this is foreseen.  Now that there's a type check, the arrays that can sneak into meta when you first initialize the data are caught.  This should be rectified separately in [pysat/pysatNASA#99](https://github.com/pysat/pysatNASA/issues/99).

aburrell: > There are too many messages :(

I could put the messages at the info level.  Since the data still exists and is loaded, that could be seen as appropriate.

rstoneback: > So, this is foreseen.

Right... I missed your note about that at the top. eek! My bad :(

> I could put the messages at the info level. Since the data still exists and is loaded, that could be seen as appropriate.

Info is definitely better, thinking maybe even debug.

rstoneback:
I also thought of a potential issue with setting new labels automagically since I did the review. If a user has a typo when trying to set an existing label they'll get a new label instead. The software will still run but the result will be unexpected from user's perspective. It is their fault, but nice if software prevents issues, where possible.

Confining new label creation to a function gets around that, but at the cost of requiring the user to know that a label is new. In practice, for users using non-default meta labels, these checks could potentially be repeated quite often. For those users, I think the hassle of additional checks favors the auto-magic option. Default meta label users may prefer option 2, it reduces potential for typos and they don't care about adding labels.

A full cost-benefit (?) here would require info on how many users are doing what with pysat, which we don't have.
 
This comment doesn't require a response, documenting my review thoughts on the topic. I'll probably think about it more.

pysat/_meta.py
aburrell: I don't want to put it at DEBUG because it would be better if people used the correct type of data, thus avoiding potential casting errors.  But I am fine with INFO, because it's just not that dire. 🐺

rstoneback: I fully agree that it is best to use the correct types of data. However, the file creation process is separated from pysat and we are unlikely to get any existing datasets changed. I thought I had the correct types for IVM but I'm getting pages of warnings every load. The log will be full of items people can't do anything about. 

I'm curious what the issue is for the IVM files. For some metadata I did pass through what I was given from earlier files but most of it was set within pysat. v2.x pysat, but still pysat.

rstoneback:
The error messages do work. 

I did a simple tweak to add existing type.
```
pysat WARNING: Metadata does not have expected type <class 'numpy.float64'><class 'numpy.int64'>. Recasting input for 'ICON_L27_Unit_Vector_Meridional_Y' with key 'Valid_Max'
pysat WARNING: Metadata does not have expected type <class 'float'><class 'numpy.int64'>. Recasting input for 'ICON_L27_Unit_Vector_Meridional_Z' with key 'ValidMax'
```
For the magnetic basis vectors I only see that warning on `ValidMax`.

Went over to OMMBV. Sure enough,
```
                              'value_min': -1.,
                              'value_max': 1}
```
typo on the `value_max` setting, copied in a number of places in the code.

aburrell:
> I also thought of a potential issue with setting new labels automagically since I did the review.

I put some thought into your comment and wanted to add further clarification to advance the discussion.  The reason I have Meta update automagically is to maintain the current behaviour, which updates the actual meta data for a data value using the dict keys regardless of the previous existence of those keys or not.  This pull request now just makes Meta self consistent.  If in the future we decide that we don't want to do this anymore, it would be a breaking change.  Perhaps it would be worth saving your thoughts as an issue for discussion/recorded decision on how flexible we want Meta to be.

rstoneback:
> > I also thought of a potential issue with setting new labels automagically since I did the review.
> 
> I put some thought into your comment and wanted to add further clarification to advance the discussion. The reason I have Meta update automagically is to maintain the current behaviour, which updates the actual meta data for a data value using the dict keys regardless of the previous existence of those keys or not. This pull request now just makes Meta self consistent. 

Thanks and good point.

> If in the future we decide that we don't want to do this anymore, it would be a breaking change. Perhaps it would be worth saving your thoughts as an issue for discussion/recorded decision on how flexible we want Meta to be.

I like the feature you added. I was definitely poking at it but am not against it. A number of pysat choices were made some time back so I try to reevaluate things as we go based on the issues we've seen in the interim.  Every design choice has strengths/weaknesses and as a reviewer I start out behind. I'm good now.

 - Since we have the check for NaNs in metadata when writing files there is protection against typos for floats. Not for strs. Int values default to -1 but since that is a valid value itself I don't think we should build a check there. It'd be nice if int data type supported nan.

rstoneback (approved):

PR #979, opened 2022-01-27
BUG: warning check

jklenzing:
pandas 1.4 throws some FutureWarnings which mess up the warning checks in test_meta.  This adjusts things so pysat does not assume the required UserWarnings are not necessarily the first warnings raised.


jklenzing:
Note there is still a potential bug if multiple UserWarnings are raised in this test, but it does not affect us at the moment.

aburrell:
Oh, and by deleting the check list you forgot to add a line to the changelog documenting this bug.

aburrell (approved):

Issue #763, opened 2021-04-08
ENH: Additional keyword support parsing filenames


rstoneback:
**Is your feature request related to a problem? Please describe.**
Current parsing on supports a fixed set of keywords. In some instrument modules the fixed keywords are repurposed. 

**Describe the solution you'd like**
Consider allowing user provided keywords. Additionally, should `inst_id` and `tag` be supported by filename parsers?

**Describe alternatives you've considered**
Leave as is.

**Additional context**
see pysatCDAAC cosmic_gps.py for an example of mapping supported keywords to items I need.

Issue #762, opened 2021-04-08
ENH: Review and document file parsing code in pysat.utils.files


rstoneback:
**Is your feature request related to a problem? Please describe.**
Filename parsing doesn't work the best if the leading and/or trailing characters are a keyword. Diagnosing this bug is challenging since a range of functions used to be one very large function.

**Describe the solution you'd like**
Iterate through the functions in order, testing each one individually. Improve docstrings, variable names, and behaviors for keywords at boundaries. In order, `construct_searchstring_from_format`, `search_local_system_formatted_filename` then `parse_fixed_width_filenames` and `parse_delimited_filenames`. The final two likely require the most attention but they each rely upon `construct_searchstring_from_format` internally.

Plus, add a note that current parsing starts at the end of the filename. 
    # When processing a fixed width filename, '?' must be used at the end so
    # the function can determine the width, but '*' may be used at the beginning
    # where information doesn't need to be parsed.

**Describe alternatives you've considered**
Leave it as is, though there are unexpected gotchas that can frustrate others.

**Additional context**

rstoneback:
Add to online documentation and docstrings as well

aburrell:
Changes in #853 may help address this issue.

jklenzing:
Linking to #854.  It is not clear from the docstrings when to use `parse_fixed_width_filenames` vs `parse_delimited_filenames`, as the example string for each is identical.

PR #982, opened 2022-01-27
ENH: Improve file parsing

rstoneback:
# Description

Addresses #763, #762

- Added support for user supplied variables in file template strings, not just those used for dates.
- Improved support for delimited filenames. More than one variable can be within a single delimited section. Delimited sections are not required to have a variable.
- Updated docstring in `from_os` as well as supporting methods. 
- Added tests for delimited filenames, user variables, and wildcard use.
- Condensed existing tests
- Output from `process_parsed_filenames` is now always sorted

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for
your test configuration

- Verified parsing delimited filenames now works on COSMIC. COSMIC filenames violate a strict interpretation of delimited information. Two examples using the `fixed_width` parser, two examples using the delimited parser.
```
pysat.Files.from_os(data_path, '{year2:04d}.{day2:03d}/ionPrf_C{sat_id:03d}.{year:04d}.{day:03d}.{hour:02d}.{minute:02d}.???_????.????_nc')
pysat.Files.from_os(data_path, '{year2:04d}.{day2:03d}/ionPrf_C{sat_id:03d}.{year:04d}.{day:03d}.{hour:02d}.{minute:02d}.G{code:02d}_{year3:04d}.{radio:04d}_nc')
pysat.Files.from_os(data_path, '{year2:04d}.{day2:03d}/ionPrf_C{sat_id:03d}.{year:04d}.{day:03d}.{hour:02d}.{minute:02d}.{code:8}.{radio:04d}_nc', delimiter='.')
pysat.Files.from_os(data_path, '{year2:04d}.{day2:03d}/ionPrf_C{sat_id:03d}.{year:04d}.{day:03d}.{hour:02d}.{minute:02d}.*.{radio:04d}_nc', delimiter='.')
```
image

 - Added unit tests

**Test Configuration**:
* Operating system: MacOS
* Version number: Python 3.9.7

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
      after adapting routine to use `Meta.rename`
    * Added an update method and type evaluation method to MetaLabels.
    * Allowed MetaLabels to be expanded through setting new Meta data values.
+   * Added support for user variables when parsing template filenames
+     in `pysat.utils.files`.
 * Deprecations
    * Removed `freq` as a standard kwarg for `pysat.Instruments.download`
    * Deprecated `_test_download_travis` as a standard attribute for
@@ -93,6 +95,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
      `utils.coords.update_longitude`
    * Fixed pysat_testmodel Instrument longitude range
    * Corrected link to testing badge in README.md
+   * `from_os` now always returns a sorted Series.
+   * Improved robustness of parsing delimited files.
 * Maintenance
    * Added unit tests for deprecation warnings related to io_utils reorg.
    * Added missing unit tests for `pysat.utils.time`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
    * Allowed MetaLabels to be expanded through setting new Meta data values.
    * Added support for user variables when parsing template filenames
      in `pysat.utils.files`.
+   * Improved robustness of parsing delimited files.
 * Deprecations
    * Removed `freq` as a standard kwarg for `pysat.Instruments.download`
    * Deprecated `_test_download_travis` as a standard attribute for
@@ -96,7 +97,6 @@ This project adheres to [Semantic Versioning](https://semver.org/).
    * Fixed pysat_testmodel Instrument longitude range
    * Corrected link to testing badge in README.md
    * `from_os` now always returns a sorted Series.
-   * Improved robustness of parsing delimited files.
 * Maintenance
    * Added unit tests for deprecation warnings related to io_utils reorg.
    * Added missing unit tests for `pysat.utils.time`
--- a/pysat/_files.py
+++ b/pysat/_files.py
@@ -912,17 +912,18 @@ class Files(object):
         part of the name that need not be extracted.
         'cnofs_cindi_ivm_500ms_{year:4d}{month:02d}{day:02d}_v??.cdf'
 
-        Leading '*' wilcards are supported when parsing fixed width filenames,
-        '*{year:4d}{month:02d}{day:02d}_v??.cdf', though the '*' is not
-        supported after the first template variable. The '?' wildcard
-        may be used anywhere in the template string.
+        When parsing using fixed width filenames (`delimiter=None`), leading
+        '*' wilcards are supported, '*{year:4d}{month:02d}{day:02d}_v??.cdf',
+        though the '*' is not supported after the first template variable. The
+        '?' wildcard may be used anywhere in the template string.
 
-        When parsing delimited filenames, the '*' wildcard is supported
+        When parsing using a delimiter, the '*' wildcard is supported
         when leading, trailing, or wholly contained between delimiters, such as
         'data_name-{year:04d}-*-{day:02d}.txt', or '*-{year:04d}-{day:02d}*',
         where '-' is the delimiter. There can not be a mixture of a template
         variable and '*' without a delimiter in between, unless the '*'
-        occurs after the variable.
+        occurs after the variable. The '?' wildcard may be used anywhere in
+        the template string.
 
         The 'day' format keyword may be used to specify either day of month
         (if month is included) or day of year.
--- a/pysat/_files.py
+++ b/pysat/_files.py
@@ -918,9 +918,11 @@ class Files(object):
         '*{year:4d}{month:02d}{day:02d}_v*.cdf' will not work.
 
         When parsing delimited filenames, the '*' wildcard is supported
-        when wholly contained between delimiters, such as
+        when leading, trailing, or wholly contained between delimiters, such as
         'data_name-{year:04d}-*-{day:02d}.txt',
         or '*-{year:04d}-{day:02d}_something.txt', where '-' is the delimiter.
+        There can not be a mixture of a template variable and '*' without a
+        delimiter in between.
 
         The 'day' format keyword may be used to specify either day of month
         (if month is included) or day of year.
--- a/pysat/_files.py
+++ b/pysat/_files.py
@@ -912,17 +912,17 @@ class Files(object):
         part of the name that need not be extracted.
         'cnofs_cindi_ivm_500ms_{year:4d}{month:02d}{day:02d}_v??.cdf'
 
-        Leading wilcards are supported when parsing fixed width filenames,
+        Leading '*' wilcards are supported when parsing fixed width filenames,
         '*{year:4d}{month:02d}{day:02d}_v??.cdf', though the '*' is not
-        supported after the first template variable. The '?' wildcard stands in place
-        of a single character and may be used anywhere in the template string.
+        supported after the first template variable. The '?' wildcard
+        may be used anywhere in the template string.
 
         When parsing delimited filenames, the '*' wildcard is supported
         when leading, trailing, or wholly contained between delimiters, such as
-        'data_name-{year:04d}-*-{day:02d}.txt',
-        or '*-{year:04d}-{day:02d}*', where '-' is the delimiter.
-        There can not be a mixture of a template variable and '*' without a
-        delimiter in between, unless the '*' occurs after the variable.
+        'data_name-{year:04d}-*-{day:02d}.txt', or '*-{year:04d}-{day:02d}*',
+        where '-' is the delimiter. There can not be a mixture of a template
+        variable and '*' without a delimiter in between, unless the '*'
+        occurs after the variable.
 
         The 'day' format keyword may be used to specify either day of month
         (if month is included) or day of year.
--- a/pysat/_files.py
+++ b/pysat/_files.py
@@ -912,6 +912,16 @@ class Files(object):
         part of the name that need not be extracted.
         'cnofs_cindi_ivm_500ms_{year:4d}{month:02d}{day:02d}_v??.cdf'
 
+        Leading wilcards are supported when parsing fixed width filenames,
+        '*{year:4d}{month:02d}{day:02d}_v??.cdf', though the '*' is not
+        supported after the first template variable, thus
+        '*{year:4d}{month:02d}{day:02d}_v*.cdf' will not work.
+
+        When parsing delimited filenames, the '*' wildcard is supported
+        when wholly contained between delimiters, such as
+        'data_name-{year:04d}-*-{day:02d}.txt',
+        or '*-{year:04d}-{day:02d}_something.txt', where '-' is the delimiter.
+
         The 'day' format keyword may be used to specify either day of month
         (if month is included) or day of year.
 
--- a/pysat/_files.py
+++ b/pysat/_files.py
@@ -920,9 +920,9 @@ class Files(object):
         When parsing delimited filenames, the '*' wildcard is supported
         when leading, trailing, or wholly contained between delimiters, such as
         'data_name-{year:04d}-*-{day:02d}.txt',
-        or '*-{year:04d}-{day:02d}_something.txt', where '-' is the delimiter.
+        or '*-{year:04d}-{day:02d}*', where '-' is the delimiter.
         There can not be a mixture of a template variable and '*' without a
-        delimiter in between.
+        delimiter in between, unless the '*' occurs after the variable.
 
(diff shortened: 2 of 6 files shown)


rstoneback:
Tests are failing due to some Meta thing which I believe @jklenzing has already identified. The pull request forms are passing.

aburrell:
The figure you use in the example doesn't show the entire command you use for testing.  Can you include that as a code snippet?

CHANGELOG.md
```diff
@@ -93,6 +95,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
      `utils.coords.update_longitude`
    * Fixed pysat_testmodel Instrument longitude range
    * Corrected link to testing badge in README.md
+   * `from_os` now always returns a sorted Series.
+   * Improved robustness of parsing delimited files.
```
aburrell, line 99: Is this improvement really a bug fix or an enhancement?

pysat/_files.py
```diff
@@ -912,6 +912,18 @@ def from_os(cls, data_path=None, format_str=None,
         part of the name that need not be extracted.
         'cnofs_cindi_ivm_500ms_{year:4d}{month:02d}{day:02d}_v??.cdf'
 
+        Leading wilcards are supported when parsing fixed width filenames,
+        '*{year:4d}{month:02d}{day:02d}_v??.cdf', though the '*' is not
```
aburrell, line 916: Make this an example that would work.

pysat/utils/files.py
```diff
@@ -116,13 +116,17 @@ def process_parsed_filenames(stored, two_digit_year_break=None):
...
         return pds.Series(None, dtype='object')
 
 
 def parse_fixed_width_filenames(files, format_str):
-    """Parse list of files, extracting data identified by format_str.
+    """Parse list of files, extracting data identified by `format_str`.
+
+    Uses the fixed-width format of the filename to find and parse
+    out variable information. The functions begins at the end of the filename,
+    parsing towards the begining.
```
aburrell, line 129: Recommend removing this, as I didn't understand what it meant until I read the 'Note' section.

pysat/utils/files.py
```diff
@@ -203,26 +227,42 @@ def parse_fixed_width_filenames(files, format_str):
 def parse_delimited_filenames(files, format_str, delimiter):
     """Parse list of files, extracting data identified by format_str.
 
+    Will parse file using `delimiter` though the function does not require
+    every delimited item to be a variable, and more than one variable
```
aburrell, line 231: Every delimited item or every parsed item?

pysat/utils/files.py
```diff
@@ -203,26 +227,42 @@ def parse_fixed_width_filenames(files, format_str):
...
+    may be within a delimited section. Thus, the main practical
```
aburrell, line 232: may exist within a parsed section?

pysat/utils/files.py
```diff
@@ -203,26 +227,42 @@ def parse_fixed_width_filenames(files, format_str):
...
+    difference with `parse_fixed_width_filenames` is more support for
+    the use of the wildcard '*' within `format_str`.
```
aburrell, line 234: Do we need the less adaptable routine, then?

pysat/utils/files.py
```diff
@@ -363,17 +411,17 @@ def construct_searchstring_from_format(format_str, wildcard=False):
...
+                for fwidth in fwidths:
                     # Make sure there is truly something there
-                    if i != 0:
+                    if fwidth != 0:
                         # Store length and add to the search string
-                        out_dict['lengths'].append(np.int64(i))
+                        out_dict['lengths'].append(np.int64(fwidth))
                         if not wildcard:
-                            out_dict['search_string'] += '?' * np.int64(i)
+                            out_dict['search_string'] += '?' * np.int64(fwidth)
```
aburrell, line 424:
```suggestion
                            out_dict['search_string'] += '?' * out_dict['lengths'][-1]
```

To avoid calling the same operation twice.

aburrell (commented):
Haven't tried things out yet, but looked over the code.

pysat/_files.py
rstoneback: That works.

pysat/utils/files.py
rstoneback: Removed.

pysat/utils/files.py
rstoneback: I think delimited, it will work on filenames that use delimiters all over the place, and not related to variables. Like, 'example-that-delimits-everything-then-quits-2009-01-03.instrument.a.cdf'

If parsed makes more sense then I'll switch.

rstoneback: All the previous tests run via the fixed_width parser are now also run using the delimited parser, even though none of those filenames were generated with delimiting in mind.

pysat/utils/files.py
rstoneback: So far there have always been filenames out there with unanticipated challenges. Possible there are some still out there.

CHANGELOG.md
rstoneback: Moved the robustness comment.

pysat/_files.py
```diff
@@ -912,6 +912,18 @@ def from_os(cls, data_path=None, format_str=None,
...
+        supported after the first template variable, thus
```
rstoneback, line 917:
```suggestion
        supported after the first template variable. The '?' wildcard stands in place
        of a single character and may be used anywhere in the template string.
```

pysat/_files.py
```diff
@@ -912,6 +912,18 @@ def from_os(cls, data_path=None, format_str=None,
...
+        '*{year:4d}{month:02d}{day:02d}_v*.cdf' will not work.
```
rstoneback, line 918:
```suggestion
```

rstoneback:
> The figure you use in the example doesn't show the entire command you use for testing. Can you include that as a code snippet?

Added.

pysat/utils/files.py
rstoneback: The fixed width parser is also less prone to false positives.

pysat/_files.py
aburrell: Not according to the words following this.  If this does work, the next sentence needs to be changed.

pysat/utils/files.py
aburrell: "The fixed width parser is also less prone to false positives."  Something like this should be added to the docstring.  It would also be good to adapt the instrument template to use this one and say in a block comment "switch to the other routine for these specific cases."

rstoneback: I added that note but then changed a function setting that reduced the potential for false positives. I changed the note  to remind the user that using the '*' too much can lead to false positives. I also updated the template instrument comment and parsing method.

pysat/_files.py
rstoneback: I reorganized the first sentence of the paragraph and such to hopefully make the description of the two different parsers clearer. I also added a comment in the delimited section about '?' which now works for that parser.

rstoneback:
Everything is breaking while the system is trying to install dependencies, in this case a version of hacking.

jklenzing:
> Everything is breaking while the system is trying to install dependencies, in this case a version of hacking.

pip is backtracking through all possible versions to determine compatibility.  Presumably this includes a pre-release version (0.6.0) where the metadata is incorrectly set.  Adding a limit to the test_requirements to skip the pre-release versions.  We may want to bump it up for consistent style checks, but this should get us over the hump for now.  The latest version is 4.0.0.

jklenzing:
the limit fixes the github actions integration.

aburrell:
Ok, can confirm this does NOT work if you use the `*` in the middle of the specifiers.  You can use a `*` on the front end, though.

pysat/utils/files.py
```diff
@@ -363,17 +438,18 @@ def construct_searchstring_from_format(format_str, wildcard=False):
...
                     # Make sure there is truly something there
-                    if i != 0:
+                    if fwidth != 0:
                         # Store length and add to the search string
-                        out_dict['lengths'].append(np.int64(i))
+                        out_dict['lengths'].append(np.int64(fwidth))
                         if not wildcard:
-                            out_dict['search_string'] += '?' * np.int64(i)
+                            val = '?' * out_dict['lengths'][-1]
+                            out_dict['search_string'] += val
```
aburrell, line 452: On line 457, recommend adding something like "this may be due to unsupported use of wildcard characters"

pysat/utils/files.py
```diff
@@ -240,47 +284,74 @@ def parse_delimited_filenames(files, format_str, delimiter):
         return stored
 
     # Parse format string to get information needed to parse filenames
-    search_dict = construct_searchstring_from_format(format_str, wildcard=True)
+    search_dict = construct_searchstring_from_format(format_str, wildcard=False)
     snips = search_dict['string_blocks']
     keys = search_dict['keys']
+    lengths = search_dict['lengths']
```
aburrell, line 290: Stylistically, is there a reason these are used instead of just using `search_dict`?  The extra variables were harder for me to track and none of the lines seem particularly long.

pysat/utils/files.py
```diff
@@ -240,47 +284,74 @@ def parse_delimited_filenames(files, format_str, delimiter):
...
-            split_name[-1] = split_name[-1].split(parsed_block[-1])[0]
-
         idx = 0
-        for isname, sname in enumerate(split_name):
-            if parsed_block[isname] != sname:
-                # Areas with data to be parsed are not equal to their
-                # `parsed_block` value
-                if stored[keys[idx]] is None:
-                    stored[keys[idx]] = [sname]
+        lsidxs = sidxs
```
aburrell, line 338: Please use a more descriptive variable name.

pysat/instruments/templates/template_instrument.py
```diff
@@ -222,21 +219,34 @@ def list_files(tag='', inst_id='', data_path=None, format_str=None):
...
 
     # We use a pysat provided function to grab list of files from the
-    # local file system that match the format defined above
-    file_list = pysat.Files.from_os(data_path=data_path, format_str=format_str)
+    # local file system that match the format defined above. This example
+    # is set to parse filenames from the local system using the delimiter '_'.
+    # Alternately, leaving the `delimiter` to the default of `None` will
+    # engage the fixed width filename parser. If there is not a common
+    # delimiter then the fixed width parser is suggested though not always
+    # required. Given the range of standards compliance across the decades of
```
aburrell, line 240:
```suggestion
    # engage the fixed width filename parser. Given the range of standards
    # compliance across the decades of
```

And fix spacing.  Recommend removing that sentence because from this level, you can't specify which parser is used.  I was reading this today and it took me a moment before I realized I had no control over that aspect.

rstoneback: A developer can control which parser is used when starting here as a template. If delimiter=None or left unassigned then the fixed width parser is used.

 I don't understand the fix spacing comment.

pysat/utils/files.py
rstoneback: Is this the error that was raised with the middle '*' character?

pysat/utils/files.py
rstoneback: Updated

pysat/utils/files.py
rstoneback: My expectation is the ValueError is reached if a user supplies a template variable without a formatting width, like '{variable:f}'

rstoneback:
> Ok, can confirm this does NOT work if you use the `*` in the middle of the specifiers. You can use a `*` on the front end, though.

You can use the '?' in the middle of specifiers though. Requires knowing ahead of time how many characters are needed.

pysat/instruments/templates/template_instrument.py
aburrell: "If there is not a common delimiter then the fixed width parser is suggested though not always required" is not accurate in this setting because you can ONLY differentiate between the two by providing or not providing a delimiter. 

"fix spacing" means that I don't know if the lines are now too long or short as a part of the whole comment block.

pysat/utils/files.py
aburrell: > Is this the error that was raised with the middle '*' character?

Yes, that's why I suggested it.

aburrell:
> Requires knowing ahead of time how many characters are needed.

Also requires that the number of characters is always the same.  For both reasons, this is why I couldn't use it.

pysat/utils/files.py
```diff
@@ -162,14 +172,19 @@ def parse_fixed_width_filenames(files, format_str):
     lengths = search_dict['lengths']
     keys = search_dict['keys']
```
aburrell, line 173: Looks like I have the same comments here about unnecessary variable names.

pysat/utils/files.py
```diff
@@ -240,47 +284,72 @@ def parse_delimited_filenames(files, format_str, delimiter):
...
-            split_name[-1] = split_name[-1].split(parsed_block[-1])[0]
-
         idx = 0
-        for isname, sname in enumerate(split_name):
-            if parsed_block[isname] != sname:
-                # Areas with data to be parsed are not equal to their
-                # `parsed_block` value
-                if stored[keys[idx]] is None:
-                    stored[keys[idx]] = [sname]
+        loop_split_idx = split_idx
```
aburrell, line 336: I know what it's for now!

pysat/instruments/templates/template_instrument.py
rstoneback: It is correct that you can only provide a single given delimiter or use the fixed width parser. In practice though, the delimited parser will likely work on filenames that don't have even a single instance of the given delimiter. I think I even test for that.  

> "fix spacing" means that I don't know if the lines are now too long or short as a part of the whole comment block.

Got it!

aburrell: GitHub shows me that this hasn't been addressed?  Which is confusing because this is marked as outdated.

rstoneback: Sometimes I think it gets confused. I updated the comment a bit to specifically note I am talking about modifying the function call `from_os` below the comment.

aburrell (approved):

Issue #947, opened 2021-10-27
ENH: epoch calculation for xarray objects


jklenzing:
**Is your feature request related to a problem? Please describe.**
pandas objects loading from netcdf4 can convert from timestamp to datetimeIndex objects.  xarray objects are left for post-processing support, handled by `pysat.instruments.methods.general.convert_timestamp_to_datetime` (eg. ICON data).

**Describe the solution you'd like**
- [x] Update `pysat.utils.io.load_netcdf_xarray` to automatically calculate DatetimeIndex appropriately -- #984
- [x] deprecate `convert_timestamp_to_datetime` -- #984
- [x] Remove deprecated function (milestone bump)

**Describe alternatives you've considered**
In pandas, the index replaces the epoch variable.  Do we replace here? Or add a new variable?  The index is required for pysat operation, but dropping a variable may be confusing.

**Additional context**
See #944

jklenzing:
Note: Make sure to incorporate `epoch_origin` and `epoch_unit` consistently so that tests for pandas can be expanded.

jklenzing:
NOTE: ICON data should use the defaults.  Make sure to check that existing code works against develop if preprocess routines are turned off.

jklenzing:
Version bump

PR #983, opened 2022-01-28
Filter metadata when loading and more!

rstoneback:
# Description

Improves the consistency and accuracy of writing files across pandas and xarray. Adds in additional features for metadata handling. I've also included bug fixes in a couple areas that were relevant. There is a feedback update in _instrument.py, I needed to modify existing feedback to diagnose where fixes were needed, then I went ahead and fixed an issue relating to that code since I was there.

In some areas I changed some default values. I've lost sense on when/where/who in some areas. I went ahead and started at the strictest level, raising a DeprecationWarning, though not all of these may actually be warranted.

Addresses https://github.com/pysat/pysatNASA/issues/99, #938, #945, #946, #974, #986

 - Adds support for filtering metadata when loading. 
   - Includes a basic filter for the metadata added by pysat when creating netCDF4 files. 
   - Also includes a hook for developers to use to construct their own metadata filtering. 
 - Adds `meta_translation` kwarg to `to_netcdf` to support translating metadata labels when writing a file.
     - Adds `meta.to_translated_dict` function which makes this possible
     - Adds default export labels for fill_val to '_FillValue', and 'FillVal'
     - Corrects `inst.generic_meta_translator` which was using older meta standard
 - Adds support for dropping metadata labels during load
 - Adds support for translation table when loading netcdf files. This is done to enable `_FillValue' in the files in backwards
   compatible way.
 - Fixes #933, metadata check applied in more instances, and with improved feedback.
 - Prevents assignment of stale data paths in stored file information.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for
your test configuration

- Testing by loading ICON IVM data and monitoring the results
- Wrote component tests for new functions
- Wrote integration tests
- Expanded number of test files written, loaded, modified, etc.
- Generated examples for documentation
- Developed jupyter notebook with a variety of file options, over in https://github.com/pysat/pysatTutorials/pull/7

**Test Configuration**:
* Operating system: MacOS
* Version number: Python 3.9.7
* Any details about your local setup that are relevant

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``main``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
[...]

PR #984, opened 2022-02-03
ENH/DEP: timestamp conversions for xarray

jklenzing:
# Description

Addresses #947

The timestamp conversion for datasets loaded using `io.load_netcdf` is not consistently handled in pandas and xarray.  This fixes that issue so there are not separate workflows for data types.  Note that `io.inst_to_netcdf` still needs to be updated.  That is documented in #991.

Converts timestamps for `xarray` datasets in a manner similar to the conversion for `pandas` datasets.  Deprecates the function `instruments.methods.general.convert_timestamp_to_datetime` as this is used by some instruments outside of the core code to convert timestamps for xarray objects.

Note that if the deprecated function `convert_timestamp_to_datetime` is run on a variable already converted to datetime, it is now passed through unchanged.  This allows for backwards compatibility for outside packaged until the deprecated function is removed.

## Type of change

# How Has This Been Tested?

Loading data from `pysatNASA`: icon_euv, icon_fuv, icon_mighti

Example:
```
import datetime as dt
import pysat
euv = pysat.Instrument('icon', 'euv')
euv.download(dt.datetime(2020, 1, 1), dt.datetime(2020, 1, 1)
euv.load(2020, 1)
euv['Epoch']
```
The data array should be `dtype='datetime64[ns]'`

- Loading data with the develop code should load as normal.  The changes here update the preprocess routine to leave data unchanged if the data is not a float.
- Comment out the preprocess line `mm_gen.convert_timestamp_to_datetime(self, sec_mult=1.0e-3)` on line 106 in the euv file.

This ensures that pysatNASA behaviour will continue across versions.

**Test Configuration**:
* Operating system: Mac OS X 11.6.3

# Checklist:

part of changelog line 35 from previous pull

### DIFF
--- a/pysat/tests/test_utils_io.py
+++ b/pysat/tests/test_utils_io.py
@@ -298,8 +298,10 @@ class TestLoadNetCDF(object):
             loaded_delta = np.diff(self.loaded_inst.index[:2])
         else:
             default_delta = np.diff(self.testInst[self.epoch_name][:2])
+
             # Average over 4 deltas to prevent rounding errors
             loaded_delta = np.diff(self.loaded_inst[test_epoch][:5]).mean()
+
         # Ratio of step_sizes should equal ratio of interpreted units
         assert ((default_delta / loaded_delta)
                 == (dt.timedelta(seconds=1) / target))
@@ -320,6 +322,7 @@ class TestLoadNetCDF(object):
             load_uts = pds.to_datetime(self.loaded_inst[test_epoch][0].values)
             default_start = (def_uts - unix_origin)
             loaded_start = (load_uts - file_origin)
+
         # Ratio of distances should equal ratio of interpreted units
         assert (default_start.total_seconds() / loaded_start.total_seconds()
                 == (dt.timedelta(seconds=1) / target))
--- a/pysat/tests/test_utils_io.py
+++ b/pysat/tests/test_utils_io.py
@@ -275,7 +275,8 @@ class TestLoadNetCDF(object):
         outfile = os.path.join(self.tempdir.name,
                                'pysat_{:}_ncdf.nc'.format(self.testInst.name))
         self.testInst.load(date=self.stime, use_header=True)
-        # TODO(#991): remove once custom epochs are integrated into `io.inst_to_netcdf for xarray
+        # TODO(#991): remove once custom epochs are integrated into 
+        # `io.inst_to_netcdf` for xarray instruments.
         if not self.testInst.pandas_format:
             # A distinct epoch name must be used for the saved xarray dataset.
             test_epoch = 'test_epoch'
--- a/pysat/tests/test_utils_io.py
+++ b/pysat/tests/test_utils_io.py
@@ -9,6 +9,7 @@ import datetime as dt
 import logging
 import numpy as np
 import os
+import pandas as pds
 import sys
 import tempfile
 import warnings
@@ -37,6 +38,7 @@ class TestLoadNetCDF(object):
         self.testInst = pysat.Instrument(platform='pysat', name='testing',
                                          num_samples=100, update_files=True)
         self.stime = pysat.instruments.pysat_testing._test_dates['']['']
+        self.epoch_name = 'time'
 
         # Create testing directory
         pysat.utils.files.check_and_make_path(self.testInst.files.data_path)
@@ -58,7 +60,7 @@ class TestLoadNetCDF(object):
         self.tempdir.cleanup()
 
         # Clear the directory attributes
-        del self.data_path, self.tempdir
+        del self.data_path, self.tempdir, self.epoch_name
         return
 
     def eval_loaded_data(self, test_case=True):
@@ -244,7 +246,8 @@ class TestLoadNetCDF(object):
         ({"epoch_name": "Santa"}, {"epoch_name": "Santa"})])
     def test_write_and_read_netcdf4_w_kwargs(self, wkwargs, lkwargs):
         """Test success of writing and reading a netCDF4 file."""
-        # Create a bunch of files by year and doy
+
+        # Create a new file based on loaded test data
         outfile = os.path.join(self.testInst.files.data_path,
                                'pysat_test_ncdf.nc')
         self.testInst.load(date=self.stime, use_header=True)
@@ -282,6 +285,14 @@ class TestLoadNetCDF(object):
         outfile = os.path.join(self.testInst.files.data_path,
                                'pysat_test_ncdf.nc')
         self.testInst.load(date=self.stime, use_header=True)
+        if not self.testInst.pandas_format:
+            # A distinct epoch name must be used for the saved xarray dataset.
+            test_epoch = 'test_epoch'
+            kwargs['epoch_name'] = test_epoch
+            # Xarray testing requires conversion to integer with ms default
+            self.testInst.data[test_epoch] = \
+                self.testInst[self.epoch_name].astype('int64') / 1.e6
+
         io.inst_to_netcdf(self.testInst, fname=outfile)
 
         # Load the data that was created
@@ -289,8 +300,13 @@ class TestLoadNetCDF(object):
         self.loaded_inst, meta = io.load_netcdf(outfile, **kwargs)
 
         # Check that the step size is expected
-        default_delta = (self.testInst.index[1] - self.testInst.index[0])
-        loaded_delta = (self.loaded_inst.index[1] - self.loaded_inst.index[0])
+        if self.testInst.pandas_format:
+            default_delta = np.diff(self.testInst.index[:2])
+            loaded_delta = np.diff(self.loaded_inst.index[:2])
+        else:
+            default_delta = np.diff(self.testInst[self.epoch_name][:2])
+            # Average over 4 deltas to prevent rounding errors
+            loaded_delta = np.diff(self.loaded_inst[test_epoch][:5]).mean()
         # Ratio of step_sizes should equal ratio of interpreted units
         assert ((default_delta / loaded_delta)
                 == (dt.timedelta(seconds=1) / target))
@@ -303,10 +319,17 @@ class TestLoadNetCDF(object):
             file_origin = unix_origin
 
         # Find distance from origin
-        default_uts = (self.testInst.index[0] - unix_origin).total_seconds()
-        loaded_uts = (self.loaded_inst.index[0] - file_origin).total_seconds()
+        if self.testInst.pandas_format:
+            default_start = (self.testInst.index[0] - unix_origin)
+            loaded_start = (self.loaded_inst.index[0] - file_origin)
+        else:
+            def_uts = pds.to_datetime(self.testInst[self.epoch_name][0].values)
+            load_uts = pds.to_datetime(self.loaded_inst[test_epoch][0].values)
+            default_start = (def_uts - unix_origin)
+            loaded_start = (load_uts - file_origin)
         # Ratio of distances should equal ratio of interpreted units
-        assert (default_uts / loaded_uts) == (dt.timedelta(seconds=1) / target)
+        assert (default_start.total_seconds() / loaded_start.total_seconds()
+                == (dt.timedelta(seconds=1) / target))
         return
 
     def test_netcdf_prevent_attribute_override(self):
@@ -376,6 +399,7 @@ class TestLoadNetCDFXArray(TestLoadNetCDF):
                                          update_files=True, num_samples=100)
         self.stime = pysat.instruments.pysat_testing2d_xarray._test_dates[
             '']['']
+        self.epoch_name = 'time'
 
         # Create testing directory
         pysat.utils.files.check_and_make_path(self.testInst.files.data_path)
@@ -403,7 +427,7 @@ class TestLoadNetCDFXArray(TestLoadNetCDF):
             pass
 
         # Clear the directory attributes
-        del self.data_path, self.tempdir
+        del self.data_path, self.tempdir, self.epoch_name
         return
 
     @pytest.mark.parametrize("kwargs,target", [({}, False),
@@ -467,6 +491,7 @@ class TestLoadNetCDF2DPandas(TestLoadNetCDF):
         self.testInst = pysat.Instrument(platform='pysat', name='testing2d',
                                          update_files=True, num_samples=100)
         self.stime = pysat.instruments.pysat_testing2d._test_dates['']['']
+        self.epoch_name = 'time'
 
         # Create testing directory
         pysat.utils.files.check_and_make_path(self.testInst.files.data_path)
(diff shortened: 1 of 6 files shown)


jklenzing:
Notes: need to update tests.  Default epoch_name is 'Epoch', test instruments all use 'time'

rstoneback:
pysat started with `Epoch` since that is what NASA uses. When I started xarray support I switched to `time` since it seemed to make more sense. We had talked about merging into one before but I think I lost the thread on that. It is possible I was supposed to do a survey. Anyway, sounds like `Epoch` vs `time` is relevant again.

jklenzing:
In that case, how about we set the default to 'time'?  It's more clear to me, and is in a uniform case.

rstoneback:
Sounds good to me.

jklenzing:
A lot of the errors happening in this pull occur now because operations are run on the epoch by default, ie, the tests fail because Santa does not exist.  🎅 😿

jklenzing:
@rstoneback, I've fixed the xarray issues, but there's a new bug popping up on the 2D pandas side:

pysat/tests/test_utils_io.py::TestLoadNetCDF2DPandas::test_read_netcdf4_w_epoch_kwargs
```
               for dim in set(two_d_dims):
                    # First or second dimension could be epoch. Use other
                    # dimension name as variable name
                    if dim[0] == epoch_name:
                        obj_key = dim[1]
                    elif dim[1] == epoch_name:
                        obj_key = dim[0]
                    else:
>                       raise KeyError('Epoch not found!')
E                       KeyError: 'Epoch not found!'

pysat/utils/io.py:502: KeyError
```
The code related to this section has not been modified.  Thoughts?

jklenzing:
One option would be to skip the pandas 2d testing here, because this test is only checking the epoch conversions, and both pandas datasets use the same code.  The full read / write is checked elsewhere.

jklenzing:
Updates from further testing:
- Running each class separately works without error.
- Running all tests in `develop` succeeds.
- Presumably something is not resetting after the xarray version of `test_read_netcdf4_w_epoch_kwargs` is run, since this is the only new set of tests run.  This contaminates the 2d pandas tests.
- I've streamlined the `setup` / `teardown` routines so we are not constantly changing user setup files. 
- I've changed the filename to be sure that the classes are not being crossed.

Out of ideas at this point.  @aburrell, any thoughts?

rstoneback:
> @rstoneback, I've fixed the xarray issues, but there's a new bug popping up on the 2D pandas side:
> 

I pushed an improvement to the error feedback that hopefully sheds some light on the issue.

rstoneback:
Looks to me when the xarray tests are run kwargs has 'epoch_name' of 'test_epoch' added. kwargs is used when loading the netCDF but not when writing, hence the disconnect. Not sure why kwargs is retaining information at the moment.

rstoneback:
Alternative path for that test is to use the epoch_name keyword argument when writing the files. I'm guessing there was an issue with that but recording it here before I forget for the weekend.

pysat/utils/io.py
```diff
@@ -716,9 +738,15 @@ def load_netcdf_xarray(fnames, strict_meta=False, file_format='NETCDF4',
...
-    # for Pandas data.
+    # If epoch can be converted to a datetime, convert
+    try:
+        data[epoch_name] = xr.DataArray(pds.to_datetime(data[epoch_name],
+                                                        unit=epoch_unit,
+                                                        origin=epoch_origin),
+                                        coords=data[epoch_name].coords)
+    except (KeyError, ValueError):
+        # Value does not exist or cannot be converted to datetime
+        pass
```
jklenzing, line 749: Reminder to self: add logger statement here with info from error for troubleshooting

jklenzing:
Thanks @rstoneback !

pysat/utils/io.py
jklenzing, line 749: Open design question: What happens if epoch data is non-convertable?
- Option A (used currently): pass and send info to the logger
- Option B: Check that data is a float / int and exists rather than use try / except.  If data is empty or already a datetime, this conversion routine is not run.  

The key difference is that sometimes a float / int that lies outside the datetime range for given epoch info will be passed through (for instance, if the user specifies a bad epoch or unit for conversion).  Do we maintain the error in this case, or leave the data unconverted and log the error as info?  Currently, this will just skip the conversion.

@aburrell, @rstoneback, any thoughts on the preferred behaviour here?

rstoneback: - First thought is the KeyError condition for epoch_name should raise an error.
 - For the ValueError. If the errors that come up are fixable by the user, like specifying the correct units, etc, then I think we should error. Make the user correct the data. 
 - If there are non-fixable things, like perhaps the file is weird, then I'm favor of using logging.warning().

aburrell: I agree with @rstoneback

jklenzing: It looks like this is a bug related to the "Santa does not exist" issue above.  `inst_to_netcdf` does not create an epoch if a random one is specified for xarray objects.  This creates the `KeyError` since this data is not contained in the datafile.  This was not discovered before since there was no check that the epoch exists in current tests.  However adding an automated calculation based on epoch having a value would error.  The "except KeyError" case was covering up a deeper bug in the system.

I propose creating an out of scope issue for this bug.  Instead of a try/except structure, check if the epoch exists, then skip the math and raise a warning to the user.  Any errors are still raised if the epoch exists and is weird.  Something like:
```
    if epoch_name in data.variables:
        data[epoch_name] = xr.DataArray(pds.to_datetime(data[epoch_name],
                                                        unit=epoch_unit,
                                                        origin=epoch_origin),
                                        coords=data[epoch_name].coords)
    else:
        pysat.logger.warning('{:} not found in datafile'.format(epoch_name))
```

The original intent of the try / except statement for ValueErrors was to skip over the conversion if the data was already a datetime object, since datasets in pysat libraries may convert separately (eg, ICON data).  However, those are done as preprocess routines, which means they are run after the load.  Additionally, they use the general library, which is updated in this pull to skip when the data is already in a datetime format.  Which means that the "except ValueError" case is not needed here.

jklenzing: As far as the unit tests go, it creates 11 errors.  I haven't seen any issues in the other packages, but could conceivably create an issue if you save an ND dataset with a custom epoch name.

```
====================================== short test summary info =======================================
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_basic_write_and_read_netcdf_mixed_case_data_format
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_basic_write_and_read_netcdf_mixed_case_meta_format
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_inst_write_and_read_netcdf - KeyErr...
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_write_and_read_netcdf4_w_kwargs[wkwargs0-lkwargs0]
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_write_and_read_netcdf4_w_kwargs[wkwargs1-lkwargs1]
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_write_and_read_netcdf4_w_kwargs[wkwargs2-lkwargs2]
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_write_and_read_netcdf4_w_kwargs[wkwargs3-lkwargs3]
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_netcdf_attribute_override - KeyErro...
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_read_netcdf4_with_time_meta_labels[kwargs0-False]
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_read_netcdf4_with_time_meta_labels[kwargs1-False]
FAILED pysat/tests/test_utils_io.py::TestLoadNetCDFXArray::test_read_netcdf4_with_time_meta_labels[kwargs2-True]
```

rstoneback: It does look like the epoch issue is unrelated to your direct code changes here. Making an issue is a reasonable way to go. I'd like to get the bug sorted out before the next release.

rstoneback: I'm going through details of file writing over in #983 so my plan has been to work the epoch bug in after that pull, or perhaps part of.

jklenzing:
Issue #991 added to capture out of scope work

pysat/tests/test_utils_io.py
```diff
@@ -278,23 +271,35 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
 
         # Check that the step size is expected
-        default_delta = (self.testInst.index[1] - self.testInst.index[0])
-        loaded_delta = (self.loaded_inst.index[1] - self.loaded_inst.index[0])
+        if self.testInst.pandas_format:
+            default_delta = np.diff(self.testInst.index[:2])
+            loaded_delta = np.diff(self.loaded_inst.index[:2])
+        else:
+            default_delta = np.diff(self.testInst[self.epoch_name][:2])
+            # Average over 4 deltas to prevent rounding errors
```
aburrell, line 301:
```suggestion
            default_delta = np.diff(self.testInst[self.epoch_name][:2])

            # Average over 4 deltas to prevent rounding errors
```

pysat/tests/test_utils_io.py
```diff
@@ -278,23 +271,35 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
+            loaded_delta = np.diff(self.loaded_inst[test_epoch][:5]).mean()
```
aburrell, line 302:
```suggestion
            loaded_delta = np.diff(self.loaded_inst[test_epoch][:5]).mean()

```

pysat/tests/test_utils_io.py
```diff
@@ -307,10 +312,17 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
-        default_uts = (self.testInst.index[0] - unix_origin).total_seconds()
-        loaded_uts = (self.loaded_inst.index[0] - file_origin).total_seconds()
+        if self.testInst.pandas_format:
+            default_start = (self.testInst.index[0] - unix_origin)
+            loaded_start = (self.loaded_inst.index[0] - file_origin)
+        else:
+            def_uts = pds.to_datetime(self.testInst[self.epoch_name][0].values)
+            load_uts = pds.to_datetime(self.loaded_inst[test_epoch][0].values)
+            default_start = (def_uts - unix_origin)
+            loaded_start = (load_uts - file_origin)
```
aburrell, line 322:
```suggestion
            loaded_start = (load_uts - file_origin)

```

pysat/tests/test_utils_io.py
```diff
@@ -278,23 +271,35 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
+            # A distinct epoch name must be used for the saved xarray dataset.
+            test_epoch = 'test_epoch'
+            kwargs['epoch_name'] = test_epoch
+            # Xarray testing requires conversion to integer with ms default
+            self.testInst.data[test_epoch] = \
+                self.testInst[self.epoch_name].astype('int64') / 1.e6
+        else:
+            if 'epoch_name' in kwargs:
+                kwargs.pop('epoch_name')
+
```
aburrell, line 288: It looks like this test needs to be different for pandas and xarray.  What do you think about breaking it into two tests and skipping if the data format is the wrong type?  Or, if there base class is pandas and an xarray class inherits this, the inhertiting class can rewrite this test with the xarray version.

aburrell (commented):
Style suggestions.  Need to see how this will affect pysatModels.

pysat/tests/test_utils_io.py
jklenzing: The difference between these is that the pandas epoch is better integrated.  This could be updated as part of #991.  Needs a todo statement.

pysat/tests/test_utils_io.py
```diff
@@ -278,23 +271,35 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
-        if not self.testInst.pandas_format:
-            pytest.skip('Not yet implemented for xarray')
-
         # Create a bunch of files by year and doy
-        outfile = os.path.join(self.testInst.files.data_path,
-                               'pysat_test_ncdf.nc')
+        outfile = os.path.join(self.tempdir.name,
+                               'pysat_{:}_ncdf.nc'.format(self.testInst.name))
         self.testInst.load(date=self.stime, use_header=True)
+        if not self.testInst.pandas_format:
```
jklenzing, line 278:
```suggestion
        # TODO(#991): remove once custom epochs are integrated into `io.inst_to_netcdf for xarray
        if not self.testInst.pandas_format:
```

pysat/tests/test_utils_io.py
```diff
@@ -278,23 +271,38 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
-        if not self.testInst.pandas_format:
-            pytest.skip('Not yet implemented for xarray')
-
         # Create a bunch of files by year and doy
-        outfile = os.path.join(self.testInst.files.data_path,
-                               'pysat_test_ncdf.nc')
+        outfile = os.path.join(self.tempdir.name,
+                               'pysat_{:}_ncdf.nc'.format(self.testInst.name))
         self.testInst.load(date=self.stime, use_header=True)
+        # TODO(#991): remove once custom epochs are integrated into `io.inst_to_netcdf for xarray
```
jklenzing, line 278:
```suggestion
        # TODO(#991): remove once custom epochs are integrated into 
        # `io.inst_to_netcdf` for xarray instruments.
```

pysat/tests/test_utils_io.py
```diff
@@ -278,23 +271,39 @@ def test_read_netcdf4_w_epoch_kwargs(self, kwargs, target):
...
-        if not self.testInst.pandas_format:
-            pytest.skip('Not yet implemented for xarray')
-
         # Create a bunch of files by year and doy
-        outfile = os.path.join(self.testInst.files.data_path,
-                               'pysat_test_ncdf.nc')
+        outfile = os.path.join(self.tempdir.name,
+                               'pysat_{:}_ncdf.nc'.format(self.testInst.name))
         self.testInst.load(date=self.stime, use_header=True)
+        # TODO(#991): remove once custom epochs are integrated into
```
jklenzing, line 278:
```suggestion
        # TODO(#991): remove once custom epochs are integrated into
```

pysat/utils/io.py
jklenzing: Dropped try/except here.  no longer needed.

jklenzing:
Merged changes from upstream.  Not sure what else is needed here beyond testing with pysatModels, etc

rstoneback:
I think this still needs changes from #983, in progress.

aburrell:
> I think this still needs changes from https://github.com/pysat/pysat/pull/983, in progress.

As of our last meeting, this wasn't a requirement for merging.  Running with pysatModels now.

aburrell (approved):
Seems to run fine with pysatModels.

PR #987, opened 2022-03-01
BUG: Instrument support module 'init' fix

rstoneback:
# Description

Addresses # (issue)

Bug fix for issue encountered reviewing the pysatMissions release candidate. Moved calls to instrument support init function to the end of pysat.Instrument instantiation to ensure the Instrument object is fully formed.

pysat.Instrument.pandas_format could be changed after instantiation but it wouldn't change any of the underlying parameters that went along with such a change. I've changed pandas_format to be a setter, so now any changes to pandas_format appropriately modify all the things that go along with it.

In perhaps a step just too far, I also added a pysat.Instrument.data_format. It can have string values of 'pandas' or 'xarray'. Added for a potential future where we have more than one supported format. Under the hood, setting pandas_format is actually setting data_format. The two attributes are the same.

*Update* Removed the additional data_format attribute. Seems like we'd need a deprecation on the pandas_format attribute and then go through the whole thing. I think that would be better suited to 3.1 or a later release. The changes for the data_format keyword are in the `data_format_addition` branch.

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for
your test configuration

- Added unit test to confirm that pysat.Instrument complete by time init() module function called.
- Added unit tests for Instrument.pandas_format and data_format (edit: removed data_format)
- Running pysatMission example 
```
   ...: import pysat
   ...: import pysatMissions
   ...: 
   ...: pysat.utils.registry.register_by_module(pysatMissions.instruments)
```

**Test Configuration**:
* Operating system: MacOS
* Version number: Python 3.9
* Any details about your local setup that are relevant - It rocks!

# Checklist:


rstoneback:
Documenting so I don't forget. This needs a test. Could do something like create an init module function that compares the Instrument passed in to a locally instantiated one, make sure all the attributes and what not are there. The Instrument init receives should be a fully complete Instrument object. It is now, but before this bug fix it wasn't.

jklenzing:
This works for pandas, but not xarray for some reason.  Both test instruments use the same init statement, though xarray has a warning thrown during init.

jklenzing:
I think we can fix this in pysatMissions rather than here to keep from breaking things elsewhere.

rstoneback:
It needs to be fixed. I know what to do, and it won’t be particularly difficult , just haven’t gotten back here yet. Good news is I’m past my hurdles of unfunded extra work and get to refocus back on code.

rstoneback:
If desired I can go through the `data_format` change now. My concern is getting already identified issues fixed before adding new ones.

pysat/_instrument.py
```diff
@@ -1973,11 +1963,36 @@ def date(self, new_date):
...
     def index(self):
         """Time index of the loaded data."""
         return self._index()
 
+    @property
+    def pandas_format(self):
+        """Boolean flag for pandas data support."""
+        return self._pandas_format
+
+    @pandas_format.setter
```
aburrell, line 1981: Recommend raising an error if attempting to update this after data has been loaded.

aburrell (changes requested):

pysat/_instrument.py
rstoneback: That is a good point. Though something like wavelet processing could trigger a type change from pandas to xarray to store the newly calculated data. We may want to leave the ability to change an existing Instrument object data type but I agree it does need something.

aburrell: > Though something like wavelet processing could trigger a type change from pandas to xarray to store the newly calculated data

This is certainly something we should consider!  But I think this is an enhancement that will require other (extensive?) changes as well.  I would recommend making that an issue, and making it impossible to do at the moment (while referencing that the error should change once the issue is addressed).

rstoneback: I just came back here to remind myself not to go adding new features! 🙈 Fortunately you were on the ball ahead of me.

Good news though, doesn't have to be an extra feature. If a user wants to convert data types via custom functions they can remove the existing data, thus making the instrument empty, change the type, and then add the new one.

My first thought for the check in the pandas_format setter is an inst.empty check. Offhand I don't remember if we have a flag that indicates if data has ever been loaded.

pysat/tests/classes/cls_instrument_property.py
```diff
@@ -706,3 +707,49 @@ def test_warn_inst_module_platform_name(self, kwargs):
 
         # Make sure isntrument loaded as inst_module
         assert tinst.inst_module == self.testInst.inst_module
+
+    def test_change_inst_pandas_format(self):
+        """Test changing `pandas_format` attribute works."""
+        new_format = not self.testInst.pandas_format
+
+        # Current data format hidden attributes
```
aburrell, line 715:
```suggestion
        # Save current data format hidden attributes
```

pysat/tests/classes/cls_instrument_property.py
```diff
@@ -706,3 +707,49 @@ def test_warn_inst_module_platform_name(self, kwargs):
...
+        current_null = self.testInst._null_data
+        current_library = self.testInst._data_library
+
+        # Assign inverted `pandas_format` setting
+        self.testInst.pandas_format = new_format
+
+        # Confirm assignment
+        assert self.testInst.pandas_format == new_format
+
+        # Confirm that internal properties have changed
```
aburrell, line 725:
```suggestion
        # Confirm assignment of visible and hidden attributes
        assert self.testInst.pandas_format == new_format
```

pysat/tests/classes/cls_instrument_property.py
```diff
@@ -706,3 +707,49 @@ def test_warn_inst_module_platform_name(self, kwargs):
 
         # Make sure isntrument loaded as inst_module
         assert tinst.inst_module == self.testInst.inst_module
+
+    def test_change_inst_pandas_format(self):
+        """Test changing `pandas_format` attribute works."""
+        new_format = not self.testInst.pandas_format
```
aburrell, line 713:
```suggestion
        new_format = not self.testInst.pandas_format
        new_type = pds.DataFrame if new_format else xr.Dataset
```

pysat/tests/classes/cls_instrument_property.py
```diff
@@ -706,3 +707,49 @@ def test_warn_inst_module_platform_name(self, kwargs):
...
+        assert not isinstance(self.testInst._null_data, type(current_null))
+        assert current_library != self.testInst._data_library
+
+        # Confirm internal consistency
+        assert isinstance(self.testInst._null_data, self.testInst._data_library)
+
+        if new_format:
+            assert isinstance(self.testInst._null_data, pds.DataFrame)
+        else:
+            assert isinstance(self.testInst._null_data, xr.Dataset)
```
aburrell, line 735:
```suggestion
        assert isinstance(self.testInst._null_data, self.testInst._data_library)
        assert isinstance(self.testInst._null_data, new_type)
```

aburrell (commented):
Made some style change suggestions to simplify the code.

aburrell (commented):
Looks good, just needs conflicts resolved.

aburrell (approved):

rstoneback:
Thanks @aburrell!

Issue #1020, opened 2022-06-13
MAINT: Remove the `use_header` kwarg


aburrell:
**Is your feature request related to a problem? Please describe.**
The introduction of `MetaHeader` required the introduction of a kwarg, `use_header`, in the `load` method to allow both the new and old load behaviour.

**Describe the solution you'd like**
- [x] Set `use_header=True` in pysat 3.2.0 release and deprecate it (so that it doesn't need to be used anymore)
- [ ] Remove `use_header` as a kwarg in pysat 3.3.0+

**Describe alternatives you've considered**
Leave the option around forever.

PR #1019, opened 2022-06-09
Header header

rstoneback:
# Description

Addresses # (issue)

Enables `use_header` at Instrument level. Warnings are reduced but there are still over 15k of them. Epoch warning would be next.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for
your test configuration

- Tweaked `use_header`. Confirmed It works as expected via terminal interaction.

**Test Configuration**:
* Operating system: MacOS
* Version number: Python 3.9
* Any details about your local setup that are relevant

# Checklist:

- [ ] Make sure you are merging into the ``develop`` (not ``main``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] Add a note to ``CHANGELOG.md``, summarizing the changes

If this is a release PR, replace the first item of the above checklist with the release
checklist on the wiki: https://github.com/pysat/pysat/wiki/Checklist-for-Release

PR #1021, opened 2022-06-13
ENH: allow `use_header` setting at init

aburrell:
Allow users to specify the `use_header` load option at initialization of an instrument instead of just when loading.
Clean up of this kwarg is now captured in #1020.

## Type of change

# How Has This Been Tested?

```
import pysat
import datetime as dt

cindi = pysat.Instrument('cnofs', 'ivm', use_header=True)
stime = dt.datetime(2009, 4, 10)
cindi.load(date=stime)
```

No deprecation warnings should appear.
**Test Configuration**:
* Version number: Python 3.8

# Checklist:

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -3123,8 +3123,9 @@ class Instrument(object):
                                    'metadata (MetaHeader). Default attachment ',
                                    'of global attributes to Instrument will ',
                                    'be Deprecated in pysat 3.2.0+. Set ',
-                                   '`use_header=True` to remove this ',
-                                   'warning.']), DeprecationWarning,
+                                   '`use_header=True` in this load call or ',
+                                   'on instrument instantiation to remove this',
+                                   ' warning.']), DeprecationWarning,
                           stacklevel=2)
             self.meta.transfer_attributes_to_instrument(self)
         self.meta.mutable = False
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -326,6 +326,12 @@ class Instrument(object):
             # Get dict of supported keywords and values
             default_kwargs = _get_supported_keywords(func)
 
+            # Expand the dict to include method keywords for load.
+            # TODO(#1020): Remove this if statement for the 3.2.0+ release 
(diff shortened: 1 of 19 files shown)


jklenzing:
Test instruments fail to load is `use_header` is not specified.  This is the line that fails, since `use_header` is not yet specified in this case:
https://github.com/pysat/pysat/blob/19f9af3d4b1e9b013a516b896d3a804bc30c728e/pysat/_instrument.py#L3115

jklenzing (commented):
I like the idea of including the option to specify `use_header` at instantiation.  There's currently a bug when the defaults are used, which should require updates in the `_instrument` code.  This fails when loading both test instruments and external instruments (ICON IVM) without specifying `use_header` (see previous comment).

Also, the deprecation warning should be updated to let users know where to specify `use_header`.
https://github.com/pysat/pysat/blob/19f9af3d4b1e9b013a516b896d3a804bc30c728e/pysat/_instrument.py#L3118-L3124

pysat/_instrument.py
```diff
@@ -1414,6 +1420,10 @@ def _load_data(self, date=None, fid=None, inc=None, load_kwargs=None):
         if load_kwargs is None:
             load_kwargs = self.kwargs['load']
 
+        # TODO(#1020): Remove after removing `use_header`
+        if 'use_header' in load_kwargs.keys():
+            del load_kwargs['use_header']
+
```
jklenzing, line 1426: not sure why this is here

jklenzing (commented):
Local loads throw warnings as expected for both test instruments and external instruments.  

We should update the tests to use this method, particularly for the deployable tests.  That could be done here or as another pull, but probably should be done for the RC for consistency

rstoneback:
I like this approach better than #1019 so I modified that one to point over here and merged in these changes. There are more places that I set `use_header` in tests over there that would be useful to pull in here. There is also a keyword at the Constellation level. If there is a potential improvement in implementation at the Constellation level as well, I'm all for it.

rstoneback (approved):

Issue #1039, opened 2022-08-25
BUG: ci_skip not working in tests


jklenzing:
**Describe the bug**
download tests that should be skipped are improperly marked in CI.  The test for download warning will be run on the ci if it should be skipped entirely.

**To Reproduce**
see pysat/pysatSpaceWeather#85

**Root Cause**
In 3.0.1:
https://github.com/pysat/pysat/blob/4d12a09ea585b88d54560413e03cae9289113718/pysat/utils/_core.py#L577-L580

If ci_skip is true, all marks should be skipped.

in 3.0.2
https://github.com/pysat/pysat/blob/3e0d33fef2a6f7a7c3d028303ddfd64477a1e48b/pysat/utils/_core.py#L470-L484

If ci_skip is true, instrument will be assumed to have no download functionality and treated appropriately.

**Route Forward**
Need to return to the original if / else structure, but keep expanded tests.  Error arose from keeping to 80 character limit without checking that logic was maintained.  Need to rename variables to fix.

jklenzing:
@rstoneback, another high-priority straightforward fix for the release candidate.

PR #1041, opened 2022-08-26
BUG: ci_skip

jklenzing:
# Description

Addresses #1039

This fixes a bug introduced in pysat 3.0.2 in the unit tests that are inherited by the ecosystem packages.  Downloads from FTP instruments are skipped in the CI environment, but run locally.  The bug mistakenly assigns the `no_download` suite of tests, which check for a warning message is downloads are not supported.

This is affecting the unit tests in pysatSpaceWeather since the release of pysat 3.0.2.  Recommend we add the fix into the RC and test a branch there against the new RC.

## Type of change

# How Has This Been Tested?

In pysat.utils._core.py, replace `(os.environ.get('CI') == 'true')` on line 470 with `(os.environ.get('CI') is None)` to simulate a True response while on a local system.  Then run:
```
import pysat
import pysatSpaceWeather
inst_list = pysat.utils.generate_instrument_list(pysatSpaceWeather.instruments)
inst_list['no_download']
```

This should provide an empty list.  When the same procedure is done in pysat develop, it lists the three FTP instruments in pysatSpaceWeather.

**Test Configuration**:
* Operating system: Mac OS X Big Sur
* Tweaks to pysat described above to simulate being in a CI environment.

# Checklist:

### DIFF
--- a/pysat/utils/_core.py
+++ b/pysat/utils/_core.py
@@ -465,7 +465,7 @@ def generate_instrument_list(inst_loc, user_info=None):
                                             inst_id=iid,
                                             temporary_file_list=True)
                     # Set flag to skip tests on a CI environment.
-                    # To test CI config on a local system, change first line
+                    # To test CI config on a local system, change first
                     # condition to (os.environ.get('CI') is None).
                     ci_skip = ((os.environ.get('CI') == 'true')
                                and not inst._test_download_ci)
--- a/pysat/utils/_core.py
+++ b/pysat/utils/_core.py
@@ -451,41 +451,45 @@ def generate_instrument_list(inst_loc, user_info=None):
                 info = {}
                 info[''] = {'': dt.datetime(2009, 1, 1)}
                 module._test_dates = info
-            for inst_id in info.keys():
-                for tag in info[inst_id].keys():
-                    inst_dict = {'inst_module': module, 'tag': tag,
-                                 'inst_id': inst_id}
+            for iid in info.keys():
+                for tag in info[iid].keys():
+                    in_dict = {'inst_module': module, 'tag': tag,
+                               'inst_id': iid}
                     # Add username and password info if needed
                     if user_info and inst_module in user_info:
-                        inst_dict['user_info'] = user_info[inst_module]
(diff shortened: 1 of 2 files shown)


pysat/utils/_core.py
```diff
@@ -451,41 +451,48 @@ def generate_instrument_list(inst_loc, user_info=None):
...
                     # Initialize instrument so that pysat can generate skip
                     # flags where appropriate
                     inst = pysat.Instrument(inst_module=module,
                                             tag=tag,
-                                            inst_id=inst_id,
+                                            inst_id=iid,
                                             temporary_file_list=True)
-                    # Set flag to skip tests on a CI environment
+                    # Set flag to skip tests on a CI environment.
+                    # To test CI config on a local system, change first line
```
jklenzing, line 468:
```suggestion
                    # To test CI config on a local system, change first
```

rstoneback:
Thanks @jklenzing. I don't really understand how the new code is different, other than variable names are updated and the conditionals in the if thens are broken out. Logically though, seems the same.

jklenzing:
> Thanks @jklenzing. I don't really understand how the new code is different, other than variable names are updated and the conditionals in the if thens are broken out. Logically though, seems the same.

In 3.0.2/3.0.3, if ci_skip is True, it will then activate the `elif not inst._password_req:` line below and apply the no_download tests (since there is no password required).  These are incorrectly applied since download support is available, just not on the CI.  Basically, the first decision is False if there is not download support (correct) or if you're in the CI environment (incorrect).  In previous iterations, it was only False if there was not download support. 

In 3.0.1 / this pull, if ci_skip is True, it skips the whole block, including the `elif not inst._password_req:` part.  Since the `if` statements were not combined previously, the `elif` line was never reached for the FTP instruments.  I've swapped the order of `if` statements here relative to 3.0.1 because the flow is more clear.   This is why I missed the logic path when I rewrote it for 3.0.2.  It's a subtle change, but an important one.

jklenzing:
The test above where you modify the code to make it think your local system is a CI environment helps show the result when you run it here vs develop.

rstoneback (approved):

PR #1069, opened 2022-12-21
STY: Clean up manual installs

jklenzing:
Looking over the logs, this is working as expected.

My thought on the netcdf4 install earlier was wrong.  Looks like the older version installs correctly after the latest fails.

Pandas has updated a new version (1.5.2) which fixes the nep29 issues, correctly installs an older version for the operational version.


aburrell (approved):

rstoneback:
Thanks @jklenzing

Issue #923, opened 2021-10-07
BUG: Loading data without a time index


rstoneback:
**Describe the bug**
If pysat gets data without time information it does not complain. While some features will continue to work, some may not, and could very well be confusing for users. 

**To Reproduce**
1. Load a netCDF file that doesn't have a time index into pysat.Instrument
2. The data is there
3. Metadata is not there
4. Some data access from Instrument will not work.

**Expected behavior**
A warning that data doesn't conform to expectations and that unexpected behavior may occur.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS 11.5.2
 - Version Python 3.8

**Additional context**

jklenzing:
Implement a warning and a switch to bypass warning

jklenzing:
Noting this issue predates the release of 3.0.2.  I think it may be stale. If I load an xarray dataset without a time index, I get an error.  Do you have an example?

aburrell:
I think the goal was to be able to load a file without a time index, so long as a date is provided instead.  As well as a bypass for the warning that gets raised with GUVI.

jklenzing:
I think the GUVI issue is separate.  That comes from the assigned epoch not being a dimension.

I'm stuck because I don't know how to test this.  If I create a xarray style dataset without a time axis, pysat already throws an error.

PR #1086, opened 2023-03-17
BUG: empty evaluation doesn't work with xarray

aburrell:
# Description

When loading data that has an xarray format, it is possible to downselect data to a non-existent time range, but still have data that doesn't depend on time in the other dimensions.  In this case, because there's data, `instrument.empty` will be False.  However, there are no index values to select.  This causes pysat to Error.

## Type of change

# How Has This Been Tested?

```
import pysat
import pysatNASA

guvi = pysat.Instrument(inst_module=pysatNASA.instruments.timed_guvi, tag='sdr-imaging', inst_id='low_res'
stime = pysatNASA.instruments.timed_guvi._test_dates[guvi.inst_id][guvi.tag]
guvi.download(start=stime)
guvi.load(date=stime, use_header=True)

```

This will error on a different branch, but work with this one.

**Test Configuration**:
* Version number: Python 3.9
* Any details about your local setup that are relevant: timed_guvi branch of pysatNASA, develop will also work.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   * Fixed `pysat.utils.io.apply_table_translation_to_file` check for duplicates
     in the meta translation table
   * Fixed an issue when passing dates through load_remote_files (#1022)
+  * Fixed a bug where data may not have any times, but still not be empty
 * Maintenance
   * Added roadmap to readthedocs
   * Improved the documentation in `pysat.utils.files`
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -3240,7 +3240,7 @@ class Instrument(object):
                 # and xarray
                 self.data = self._next_data.copy()
                 self.data = self[temp_time:last_pad]
-                if not self.empty:
+                if len(self.index) > 0:
                     if (self.index[0] == temp_time):
                         self.data = self[1:]
                     self.concat_data(stored_data, prepend=True)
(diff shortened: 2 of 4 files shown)


jklenzing:
@aburrell, The test snippet above works on the develop branch.  I'm not sure what I'm supposed to be looking for.

aburrell:
I am not sure then, it didn't work for me when I was working on the updated timed_guvi pull.  I also created a test that would cause the code to fail, so I am pretty sure the problem exists.

jklenzing:
OK, it works with pysatNASA develop, but not the new timed_guvi pull.

The root cause comes from our assumptions about what an empty xarray dataset looks like:
https://github.com/pysat/pysat/blob/34b9125b86ef93adf78f3041d29fde8b5b014fd7/pysat/_instrument.py#L1222-L1233

Not sure if we should modify this instead or leave it at the broken function since that requires an index.  The instrument file calls `self.empty` a lot, which could mean we have a lot of other potential bugs, or we end up breaking things more if we require an index.

aburrell:
Yeah, I wasn't sure either.  We're so dependent on an index that I think we would be safe to say "even if there's other dimensions, there's no time dimension so that means there's no data".  That wouldn't be the fix I did here, but it would be a more stringent and pysat-entric fix.  I am more happy to go that way if there's a consensus that it's appropriate.

jklenzing:
I think in light of the discussion on #923, we should use the solution you've implemented here.  The route for that issue is to warn in there is no time index.

jklenzing (approved):

PR #1095, opened 2023-03-29
ENH/BUG: added a function to update fill values

aburrell:
# Description

Potentially addresses #707 by adding a core utility to consistently update fill values in the metadata and data simultaneously.  Also fixes issues found when testing that this worked as desired when printing netCDF files.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

```
import pysat

inst = pysat.Instrument('pysat', 'testing', use_header=True)
inst.load(2009, 1)

print(inst.meta['mlt', inst.meta.labels.fill_val])  # Should be np.nan
pysat.utils.update_fill_values(inst, 'mlt', -1.0)
print(inst.meta['mlt', inst.meta.labels.fill_val])  # Should be -1.0
```

**Test Configuration**:
* Operating system: OS X Big Sur
* Version number: Python 3.9
* Any details about your local setup that are relevant

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``main``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] Add a note to ``CHANGELOG.md``, summarizing the changes

If this is a release PR, replace the first item of the above checklist with the release
checklist on the wiki: https://github.com/pysat/pysat/wiki/Checklist-for-Release

PR #1108, opened 2023-04-08
BUG: setting meta data type

jklenzing:
# Description

Follows #1095

When data is set to new values, the meta types are updated accordingly.  However, in some cleaning routines, values are updated to `np.nan`.  This breaks the type update.

~~This bugfix only updates the type if the value is non a nan.~~

The source of the bug is that `meta._data_types` is initiated as `None` by default.  When new data is set, keys cannot be added to a `NoneType`

## Type of change

# How Has This Been Tested?

By running pytest on pysatNASA.

**Test Configuration**:
* Operating system: Monterrey
* Version number: Python 3.9.7

# Checklist:

- [ ] Add a note to ``CHANGELOG.md``, summarizing the changes -- bugfix to last pull

### DIFF
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -1003,9 +1003,8 @@ class Instrument(object):
                     # (including list or slice).
                     self.data.loc[self.data.index[key[0]], key[1]] = new
 
-                if not np.isnan(new):
-                    self.meta._data_types[key[1]] = self.data[
-                        key[1]].values.dtype.type
+                self.meta._data_types[key[1]] = self.data[
+                    key[1]].values.dtype.type
                 self.meta[key[1]] = {}
                 return
             elif not isinstance(new, dict):
--- a/pysat/_instrument.py
+++ b/pysat/_instrument.py
@@ -1003,8 +1003,9 @@ class Instrument(object):
                     # (including list or slice).
                     self.data.loc[self.data.index[key[0]], key[1]] = new
 
-                self.meta._data_types[key[1]] = self.data[
-                    key[1]].values.dtype.type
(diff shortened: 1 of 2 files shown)


pysat/_instrument.py
```diff
@@ -1003,8 +1003,9 @@ def __setitem__(self, key, new_data):
                     # (including list or slice).
                     self.data.loc[self.data.index[key[0]], key[1]] = new
 
-                self.meta._data_types[key[1]] = self.data[
-                    key[1]].values.dtype.type
+                if not np.isnan(new):
+                    self.meta._data_types[key[1]] = self.data[
+                        key[1]].values.dtype.type
```
aburrell, line 1008:
```suggestion
                if np.isnan(new):
                    self.meta._data_types[key[1]] = float
                else:
                    self.meta._data_types[key[1]] = self.data[
                        key[1]].values.dtype.type
```

Is this a better way to move forward?

aburrell (commented):
A question on the approach.

pysat/_instrument.py
jklenzing: I still get the same error when new is a nan:
```
TypeError: 'NoneType' object does not support item assignment
```

jklenzing: Actually, if I change the set function to 37.0 rather than nan, I still get the same error.

jklenzing: Found the source of the bug deeper in.  Updates pending the local tests...

jklenzing:
The solution is more complex than I thought.  Closing in favor of writeup in #1109

PR #1131, opened 2023-07-20
ENH: allow custom `concat_data` methods

aburrell:
# Description

Addresses pysatNASA issue https://github.com/pysat/pysatNASA/issues/197 by allowing custom concatenation methods to be supplied by Instruments.  This is necessary when there are multiple time indices in a file, or data without time indices.

## Type of change

# How Has This Been Tested?

Use with pysatNASA pull request: https://github.com/pysat/pysatNASA/pull/198

**Test Configuration**:
* Version number: Python 3.9
* Any details about your local setup that are relevant: https://github.com/pysat/pysatNASA/pull/198

# Checklist:


aburrell:
@jklenzing I couldn't figure how how/where to put a test that ensures the extra time dimensions have the same max/min range as the main time dimension.  Some guidance/support there would be appreciated.

pysat/instruments/pysat_ndtesting.py
```diff
@@ -17,6 +17,8 @@
 tags = {'': 'Regular testing data set'}
 inst_ids = {'': ['']}
 _test_dates = {'': {'': dt.datetime(2009, 1, 1)}}
+_test_load_opt = {'': {'': [{'num_extra_time_coords': 0},
+                            {'num_extra_time_coords': 1}]}}
```
aburrell, line 21: Drop in coverage seems to indicate that these aren't being used 🐛

jklenzing:
@aburrell, plugging in the test code
```
import pysat
test = pysat.Instrument('pysat', 'ndtesting', use_header=True, num_extra_time_coords=1)
test.load(2009, 1, 2009, 3)
```
Loads two days of data as expected. However, any print lines I put in the `concat_data` routine are not printed.

The test code up in #1132 seems to be running ok, but the same lines in the method show up as uncovered.

aburrell:
@jklenzing can you tell if the load option is being called?  Is there a reason for concat to use the other lines?

aburrell:
The big problem was a gap in the `test_instruments` testing suite, which I plugged with another load test.

jklenzing:
> The big problem was a gap in the `test_instruments` testing suite, which I plugged with another load test.

We currently only download one day of data, so there's nothing to pad, although the routines will still probably run.

aburrell:
With your new addition we'll download one more day, no?

aburrell:
It also does call the concat and run through the padding lines for the instruments, which is probably a good thing since that's where I discovered the bug.

jklenzing:
> With your new addition we'll download one more day, no?

Yeah, the other pull should address that.  Finding that a bunch of instruments don't concat properly though.

aburrell:
@jklenzing the 3.6 tests are failing probably because of array size.  I made the ndtesting x,y,z coordinates smaller.  Are there other ways to make that data set smaller?

jklenzing:
We could change the time cadence potentially

aburrell:
Ok, that didn't work so now I am out of ideas 😿 I might have more ideas later.

aburrell:
@jklenzing all the tests failing here are passing locally.  Can you take a look at the tests failing on the non-3.6 versions?

jklenzing:
> @jklenzing all the tests failing here are passing locally. Can you take a look at the tests failing on the non-3.6 versions?

Tests pass locally with xarray 2023.6.0, the same ones fail with 2023.8.0 (the version GA is using)

jklenzing:
I suspect that lines 671-688 in the constellation file will need to be updated for the new coord usage (text from https://docs.xarray.dev/en/stable/whats-new.html below).

[Coordinates](https://docs.xarray.dev/en/stable/generated/xarray.Coordinates.html#xarray.Coordinates) can now be constructed independently of any Dataset or DataArray (it is also returned by the [Dataset.coords](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.coords.html#xarray.Dataset.coords) and [DataArray.coords](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.coords.html#xarray.DataArray.coords) properties). Coordinates objects are useful for passing both coordinate variables and indexes to new Dataset / DataArray objects, e.g., via their constructor or via [Dataset.assign_coords()](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.assign_coords.html#xarray.Dataset.assign_coords). We may also wrap coordinate variables in a Coordinates object in order to skip the automatic creation of (pandas) indexes for dimension coordinates. The [Coordinates.from_pandas_multiindex](https://docs.xarray.dev/en/stable/generated/xarray.Coordinates.from_pandas_multiindex.html#xarray.Coordinates.from_pandas_multiindex) constructor may be used to create coordinates directly from a [pandas.MultiIndex](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.MultiIndex.html#pandas.MultiIndex) object (it is preferred over passing it directly as coordinate data, which may be deprecated soon). Like Dataset and DataArray objects, Coordinates objects may now be used in [align()](https://docs.xarray.dev/en/stable/generated/xarray.align.html#xarray.align) and [merge()](https://docs.xarray.dev/en/stable/generated/xarray.merge.html#xarray.merge). ([GH6392](https://github.com/pydata/xarray/issues/6392), [PR7368](https://github.com/pydata/xarray/pull/7368)). By [Benoît Bovy](https://github.com/benbovy).

pysat/instruments/pysat_ndtesting.py
```diff
@@ -26,13 +28,15 @@
...
-# Optional method, preprocess
+# Optional methods
+concat_data = mm_test.concat_data
 preprocess = mm_test.preprocess
 
 
 def load(fnames, tag='', inst_id='', non_monotonic_index=False,
          non_unique_index=False, malformed_index=False, start_time=None,
-         num_samples=864, test_load_kwarg=None, max_latitude=90.):
+         num_samples=96, test_load_kwarg=None, max_latitude=90.0,
```
aburrell, line 38:
```suggestion
         num_samples=864, test_load_kwarg=None, max_latitude=90.0,
```

pysat/instruments/pysat_ndtesting.py
```diff
@@ -58,14 +62,16 @@ def load(fnames, tag='', inst_id='', non_monotonic_index=False,
         will begin at midnight. (default=None)
     num_samples : int
         Maximum number of times to generate.  Data points will not go beyond the
-        current day. (default=864)
+        current day. (default=96)
```
aburrell, line 65:
```suggestion
        current day. (default=864)
```

pysat/tests/classes/cls_instrument_library.py
```diff
@@ -512,6 +513,72 @@ def test_clean_warn(self, clean_level, inst_dict, caplog):
...
+            Valid pad value for initializing an instrument
+        inst_dict : dict
+            Dictionary containing info to instantiate a specific instrument.
+            Set automatically from instruments['download'] when
+            `initialize_test_package` is run.
+
+        """
+        # Skip for Python 3.6, keeping information that will allow adding
+        # or skipping particular instruments.
+        # TODO(#1136): Remove skip once memory management is improved
```
aburrell, line 537: Fixing this test in Python 3.6 will require a major overhaul in pysat, and so cannot be tackled at the moment.  Recorded here: https://github.com/pysat/pysat/issues/1136

aburrell:
@jklenzing this is now ready to review.

pysat/_instrument.py
```diff
@@ -788,20 +791,22 @@ def __getitem__(self, key):
             inst[datetime1:datetime2, 'name1':'name2']
 
         """
+        if data is None:
+            data = self.data
```
JonathonMSmith, line 795: When would you want to select from another data object, from meta? Or is there another case for this. (I like this change btw)

pysat/_instrument.py
```diff
@@ -3240,14 +3331,16 @@ def load(self, yr=None, doy=None, end_yr=None, end_doy=None, date=None,
                 self._next_data = getattr(self._next_data,
                                           sort_method)(*sort_args)
 
-            # Make tracking indexes consistent with new loads
+            # Make tracking indexes consistent with new loads, as date loading
+            # and file loading have to be treated differently due to change in
+            # inclusive/exclusive range end treatment. Loading by file is
+            # inclusive.
             if self._load_by_date:
+                # Arithmatic uses datetime or DateOffset objects
```
JonathonMSmith, line 3339:
```suggestion
                # Arithmetic uses datetime or DateOffset objects
```

pysat/_instrument.py
aburrell: The case that led to this change is in `load`.  When you pad data, you want to the previous and next days, select the appropriate time window, and then concat the data together.  Without this, you couldn't slice reliably without reassigning the data you want to slice to `self.data`.  Now you can!

pysat/_instrument.py
```diff
@@ -3287,48 +3380,40 @@ def load(self, yr=None, doy=None, end_yr=None, end_doy=None, date=None,
...
+                ndata = self.__getitem__(slice(self.index[-1], last_pad),
+                                         data=self._next_data)
+                if not self._empty(ndata):
+                    # Test the data index, slicing if necessary
+                    nindex = self._index(data=ndata)
+                    if nindex[0] == self.index[-1]:
+                        ndata = self.__getitem__(
+                            slice(1, len(ndata[self.index.name])), data=ndata)
+                    self.concat_data(ndata, prepend=False)
+                del ndata
```
aburrell, line 3408: @JonathonMSmith here's the bit of code that I was talking about.  See how much simpler it is now!

JonathonMSmith: The old code mentions in the comments that __getitem__ is used but I don't see it used explicitly, was it being used at all before?
Also the boolean for the concat operations have changed. Is that because of all the copying that was going on before. It's much more intuitive that the previous data would be prepended... instead of appending the current data to the previous data lol

JonathonMSmith (approved):

pysat/_instrument.py
aburrell: Yes, `__getitem__` is a magic function, which means it's present in all classes and gets data from the class.  However, it's not meant to be called by a user as a method or function.  So when you do something like `inst['varname']` to get a particular variable from an Instrument, the Instrument class is actually calling `__getitem__`.  When you have a complicated class like `Instrument`, you can write your own `__getitem__` so long as you follow the rules for what this method requires.  In the old code `self.data[:-1]` was actually calling `__getitem__`.

You're right that all the copying going on meant that we needed to switch the way concat combined the data.  Now it's more straightforward.

Issue #1118, opened 2023-04-25
STY: space after operator (E275)


jklenzing:
**Is your feature request related to a problem? Please describe.**
`pycodestyle` 2.9.1 changes how E275 is checked. This creates an error where `if(~~~)` is used for lengthy evaluation statements so that the tab for a continued line is not 4 characters.  This style issue pops up 15 times in the code.

***EDIT:*** Looking over the rationale at `pycodestyle`, the intent looks like the whitespace should distinguish between a statement (if, assert, raise) and a function.

**Describe the solution you'd like**
Update syntax accordingly. For example:
```
if(not isinstance(self.file_format, str)
   or (self.file_format.find("{") < 0)
   or (self.file_format.find("}") < 0)):
```
becomes
```
cond1 = not isinstance(self.file_format, str)
cond2 = self.file_format.find("{") < 0
cond3 = self.file_format.find("}") < 0
if (cond1 or cond2 or cond3):
```
or potentially
```
if (any([not isinstance(self.file_format, str),
         (self.file_format.find("{") < 0),
         (self.file_format.find("}") < 0)]):
```
**Describe alternatives you've considered**
Ignore E275 and leave the code alone

**Additional context**
Discovered during #1116.  Set to ignore this rule there, but should be a larger discussion among developers on how to move forward.

jklenzing:
It's fine like that.  But then the code run conditionally follows the same indentation level as the continued conditions.

jklenzing:
> Looking at [rule E275](https://www.flake8rules.com/rules/E275.html), I'm not sure where the error is coming from here

I think it's just the missing whitespace

jklenzing:
> It's fine like that. But then the code run conditionally follows the same indentation level as the continued conditions.

that sentence got away from me somewhere...

jklenzing:
Found the cause: it appears to be an intentional change with `pycodestyle` as of version 2.9.0: https://github.com/PyCQA/pycodestyle/issues/1096

jklenzing:
Pinging @aburrell, any thoughts on which option to pursue? I'm leaning toward using `any` and `all` to bump the indentation point over

aburrell:
Ugg, I find the change annoying (probably because it's a change).  What's the execution time difference on them?

jklenzing:
Running code snippets in `timeit`, tiny.

aburrell:
I suppose using `any` and `all` is going to make the line breaks easier, so doing that is probably the least painful thing.

PR #1144, opened 2023-10-06
STY: fix whitespace checks

jklenzing:
# Description

Addresses #1118

pycodestyle has updated a bug when scanning for E275 (whitespace). `if(` is no longer allowed, as this treats `if` as a function rather than an operator. This has been used previously when dealing with long lines that are continued to distinguish the condition from the code run.

Updates usage throughout, primarily using `any` and `all` to replace `or` and `and` to make tab alignment work out.. Nested `if` statements are used when needed since later statements may not run correctly for certain data types. (ie, `func(a) and func(b)`, where `func(b)` produces an error where `func(a)` is False.)

This allows us to remove the cap on `hacking`, which sets other version limits including `pycodestyle`.

## Type of change

# How Has This Been Tested?

pytest

**Test Configuration**:
* Operating system: Monterrey
* Version number: Python 3.10.9

# Checklist:


pysat/_instrument.py
```diff
@@ -409,11 +409,17 @@ def __init__(self, platform=None, name=None, tag='', inst_id='',
...
+                    raise ValueError(''.join(['file format set to default, ',
+                                              'supplied string must be ',
+                                              'iterable [{',
+                                              self.file_format, '}]']))
+            else:
                 raise ValueError(''.join(['file format set to default, ',
-                                          'supplied string must be iterable ',
+                                          'supplied format must be iterable ',
+                                          'string',
                                           '[{:}]'.format(self.file_format)]))
```
jklenzing, line 423: The first check is for non-string vals , which must be done separately (as in repeated `or` statements). Since these need to be broken out, adjusting error message for non-string inputs.

pysat/_instrument.py
```diff
@@ -409,11 +409,17 @@ def __init__(self, platform=None, name=None, tag='', inst_id='',
...
-            if(not isinstance(self.file_format, str)
-               or (self.file_format.find("{") < 0)
-               or (self.file_format.find("}") < 0)):
+            if isinstance(self.file_format, str):
+                if any([self.file_format.find("{") < 0,
+                        self.file_format.find("}") < 0]):
+                    raise ValueError(''.join(['file format set to default, ',
+                                              'supplied string must be ',
+                                              'iterable [{',
+                                              self.file_format, '}]']))
```
aburrell, line 418:
```suggestion
                    raise ValueError(''.join(['file format must be an ',
                                              'iterable string with key formatting [{',
                                              self.file_format, '}]']))
```
Or something that makes more sense

aburrell (commented):
One of the messages needs to be fixed to be more informative

pysat/_instrument.py
```diff
@@ -409,11 +409,18 @@ def __init__(self, platform=None, name=None, tag='', inst_id='',
         # Check to make sure value is reasonable
         if self.file_format is not None:
             # Check if it is an iterable string
-            if(not isinstance(self.file_format, str)
-               or (self.file_format.find("{") < 0)
-               or (self.file_format.find("}") < 0)):
+            if isinstance(self.file_format, str):
+                if any([self.file_format.find("{") < 0,
+                        self.file_format.find("}") < 0]):
+                    raise ValueError(''.join(['file format set to default, ',
```
aburrell, line 415: This will also trigger if the supplied string is not the default, but doesn't have any '{}' specifiers.

jklenzing: That is the intention, unless I misunderstand

aburrell: Yes, it is.  But the message doesn't tell the user that's the problem.  It tells the user it's a different problem, the "default" is now in a separate error (the `else` below).

jklenzing: The if/else here is for string input or non-string input, not non-default.  Though since it's an error message, it can't set back to default, so this is confusing

aburrell (approved):
New uncovered lines are just the re-written lines that were already not covered.

PR #1175, opened 2024-01-31
Update main.yml

jklenzing:
🤷

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
-        python-version: ["3.10", "3.11"]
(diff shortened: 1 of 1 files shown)


jklenzing:
Getting one consistent error in python 312 across all versions.
```
 FAILED pysat/tests/test_utils.py::TestFmtCols::test_fmt_raises[max_num--10-ValueError-max() arg is an empty sequence] - AssertionError: unexpected error message ('max() arg is an empty sequence' not in 'max() iterable argument is empty')
```

aburrell:
> Getting one consistent error in python 3.12 across all versions.

Sounds like a good issue for either the current or the next release, depending on how simple it is 🐛

jklenzing:
Looking over the code, I think this is a matter of python updating the error message. We would either have to 
- change the error message we check based on python version
- use a simpler string that is captured in both, e.g, 'empty'

aburrell (approved):

PR #1224, opened 2026-04-03
Add python3.14 and perform maintenance

aburrell:
# Description

Addresses deprecations that have crept in from the last release.  Fixes #1223 by removing a version cap in the docs requirements and #974 by removing support of Python 3.9.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update

# How Has This Been Tested?

Built docs and ran unit tests locally.

**Test Configuration**:
* Operating system: OS X Tahoe
* Version number: Python 3.10
* Any details about your local setup that are relevant

# Checklist:

- [x] Make sure you are merging into the ``develop`` (not ``main``) branch
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have linted the files updated in this pull request
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] Add a note to ``CHANGELOG.md``, summarizing the changes
- [x] Update zenodo.json file for new code contributors

If this is a release PR, replace the first item of the above checklist with the
release checklist on the wiki:
https://github.com/pysat/pysat/wiki/Checklist-for-Release

PR #1225, opened 2026-06-01
Add python3.14v2

### LINKED ISSUE #974: STY: Windows and temporary direcotories
**Is your feature request related to a problem? Please describe.**
Windows has issues sometimes cleaning up temporary directories.  Multiple errors including `PermissionError` are thrown at the same time. This largely affects how unit tests are run in the CI environment. Python 3.10 fixes this by adding the  `ignore_cleanup_errors` kwarg.

**Describe the solution you'd like**
Update usage of `tempfile.TemporaryDirectory` to include `ignore_cleanup_errors=True`.  In the meantime, a `try` / bare `except` is used around cleanup statements where known problems occur to keep the code running until older python support can be dropped (estimated to be sometime around the release of Python 3.12).

**Describe alternatives you've considered**
🚫 :window:

**Additional context**
See discussion in #966

rstoneback:
# Description

Addresses #(issue)

Continues the pull request started in #1224 and addresses remaining bits including new reported changes to xarray.

## Type of change

# How Has This Been Tested?

- Online and local tests which will be continued

**Test Configuration**:
* Operating system: 13.X
* Version number: Python 3.X

# Checklist:

- [x] I have linted the files updated in this pull request
- [x] Update zenodo.json file for new code contributors

If this is a release PR, replace the first item of the above checklist with the
release checklist on the wiki:
https://github.com/pysat/pysat/wiki/Checklist-for-Release


rstoneback:
First, establishing a testing baseline.

pysat/_meta.py
```diff
@@ -398,9 +398,16 @@ def __setitem__(self, data_vars, input_dat):
                     good_set = True
 
                     # See if this meta data key has already been defined
-                    # in MetaLabels
+                    # in MetaLabels, either as the label or the label value
+                    iattr = None
                     if ikey in self.labels.label_attrs.keys():
                         iattr = self.labels.label_attrs[ikey]
+                    elif ikey in self.labels.label_attrs.values():
```
rstoneback: OK. So Meta is currently setup for things like, 
```
        meta['var_name'] = {meta.labels.name: 'Variable Name',
                            meta.labels.units: 'MegaUnits'}
```
where 'meta.labels.name' resolves to a file specific value, like 'long_name', but this code change would allow:
```
        meta['var_name'] = {'name': 'Variable Name',
                            'units': 'MegaUnits'}
```
to also set the metadata, where 'name' is translated internally to 'long_name'.

@aburrell can you provide some context as to why you are interested in this change? 

Because both values and keys are supported there may be a future user issue where they meant one but got the other. Like, what if for weird reasons meta.labels.units resolves to 'name'. A user could assign metadata using 'name' expecting it to go to the meta.labels.names area but instead sets the units area.

aburrell: I don't really care, do what you like so long as the tests pass.

rstoneback: ok, then I'm going to remove this part as we don't need it for a maintenance release.

pysat/tests/test_utils_io.py
```diff
@@ -303,14 +302,18 @@ def test_inst_write_and_read_netcdf(self, add_path):
 
         del netcdf_inst.data, netcdf_inst
 
-        # TODO(#974) It appears the source of the open references is related
```
rstoneback, line 306: So I tried creating a test that would let us know in the future if windows and files ever get sorted out. Unfortunately, the code below didn't quite work as expected. My expectation is if the except area isn't triggered the else would run, but that isn't what we got. Anyway, the real story was getting our tests running right, not sorting out windows. In this process I did find an area that didn't have a temporary directory cleanup, and I also confirmed that we are properly dealing with errors from windows and temp files/directories. 
```
        try:
            os.remove(outfile)
        except Exception as err:
            # Problems expected in windows
            if platform.system() != 'Windows':
                raise err
        else:
            # When problem is fixed in windows
            if platform.system() == 'Windows':
                raise OSError("Windows can correctly manage files! See #974")
```

pysat/_meta.py
rstoneback: Ah, I get it now that I see the errors without the new meta code. I think it is an error in some of the tests. Tracking it down.

rstoneback:
The issues with Meta appear sorted out. Test failures are now related to running multiple instances of pysat at the same time....

pysat/tests/test_utils_io.py
rstoneback: So I added an updated version of the code above to ensure the else code was only running if there was no error on windows. It gets triggered, implying no exception. However if I run a bare os.remove it produces an exception. So this is the windows version of Schrödinger's cat, no matter what it produces an error. Taking this out again. Core windows problems aren't actually our problem.

pysat/utils/_core.py
```diff
@@ -615,17 +615,21 @@ def update_fill_values(inst, variables=None, new_fill_val=np.nan):
                 # Update the Meta data
                 inst.meta[var] = {inst.meta.labels.fill_val: new_fill_val}
 
-                # Update the variable data
+                # Update the variable data, masking is much faster than
+                # indexing
```
rstoneback: Can you provide a source for masking being faster? What I've seen is indexing is generally faster. https://stackoverflow.com/a/46043964

rstoneback:
One or fewer instances can generate intermittent failures from `TestFileDirectoryTranslations.test_updating_directories`. Rerunning the tests usually clears the error however coveralls doesn't like to accept results from rerunning tests, which creates its own error. Poking at it now.

.github/workflows/pip_rc_install.yml
```diff
@@ -13,30 +13,30 @@ jobs:
...
+            pandas_ver: "2.3.0"
+            scipy_ver: "1.14.0"
+            xarray_ver: "2024.5.0"
             os: ubuntu-latest
             test_config: "SPEC0000"
           # Operational compliance settings
-          - python-version: "3.9"
-            numpy_ver: "1.23.5"
+          - python-version: "3.12"
+            numpy_ver: "2.0.0"
```
rstoneback, line 30: We have listed minimum versions of packages. Seems like they should go here. We list 1.19 as minimum numpy version.

rstoneback:
Tests passed but a couple failed due to network stuff. The second time the tests themselves passed but Coveralls throws up an error because test results can't be updated. I'll try one more time but back issues are increasing.

rstoneback:
Hey @namurphy do you still happen to have your test setup available? I can't get GitHub actions to install the minimum versions you found in your previous PR. We've had a few code changes since then so I'm unsure if those minima are still good. Would it be possible for you to recheck those tests? If not, I can move forward with the versions being tested here on GitHub. Thanks!

pysat/utils/_core.py
```diff
@@ -615,17 +615,35 @@ def update_fill_values(inst, variables=None, new_fill_val=np.nan):
                 # Update the Meta data
                 inst.meta[var] = {inst.meta.labels.fill_val: new_fill_val}
 
-                # Update the variable data
                 try:
                     if np.isnan(old_fill_val):
+                        # Needed for NaNs
```
rstoneback, line 620: I had to modify the mask approach since it didn't work with older xarray versions

rstoneback:
> Tests passed but a couple failed due to network stuff. The second time the tests themselves passed but Coveralls throws up an error because test results can't be updated. I'll try one more time but back issues are increasing.

Some good news here is I found a fix from coveralls. Working so far. https://github.com/lemurheavy/coveralls-public/issues/1716

pysat/tests/test_files.py
```diff
@@ -1208,27 +1209,11 @@ def create_instrument(j):
     update_files should update the file list in .pysat
 
     """
-    root_fname = ''.join(('pysat_testing_junk_{year:04d}_{month:02d}',
```
rstoneback, line 1211: This code area is the original code. Later separated into setup and teardown methods, making this stuff here redundant. Removed.

pysat/tests/test_meta.py
```diff
@@ -441,13 +441,13 @@ def test_equality(self):
         cmeta = pysat.Meta()
 
         for mobj in [self.meta, cmeta]:
-            mobj['test_var'] = {'units': 'testU',
```
rstoneback, line 444: Old style format that was missed when moving to Meta.labels. That is my guess on the origin but regardless the style is out of date and causing problems.

pysat/utils/_core.py
```diff
@@ -615,17 +615,35 @@ def update_fill_values(inst, variables=None, new_fill_val=np.nan):
...
+                        ifill = ifill[0]
+                        inst[ifill, var] = new_fill_val
+                    else:
+                        # Multidimensional xarray
+                        # I think the most correct solution would be to use
+                        # inst[*ifill, var] = new_fill_val
+                        # which works, but takes forever for some reason on
+                        # at least one test
+                        # (test_update_fill_values_by_type[testmodel]).
+                        # TODO(#1226) Sort out why *ifill is lacking performance
```
rstoneback, line 645: Initial check is that we first try to assign by value but then fall back to integer indexing. Initial evidence suggests it is the xarray value check that is taking the longest.

.github/workflows/pip_rc_install.yml
```diff
@@ -13,30 +13,30 @@ jobs:
...
+            pandas_ver: "2.3.0"
+            scipy_ver: "1.14.0"
+            xarray_ver: "2024.5.0"
             os: ubuntu-latest
             test_config: "SPEC0000"
           # Operational compliance settings
-          - python-version: "3.9"
-            numpy_ver: "1.23.5"
+          - python-version: "3.12"
+            numpy_ver: "1.19"
```
aburrell:
```suggestion
            numpy_ver: "2.0.0"
```

pysat/tests/test_meta.py
aburrell: No, this test was kept as is to ensure string inputs could be used as meta input.  This is removing a test functionality.

pysat/utils/files.py
```diff
@@ -826,8 +826,9 @@ def update_data_directory_structure(new_template, test_run=True,
                         nold = len(inst.files.files)
                         if not test_run:
                             if nnew != nold:
-                                estr = ' '.join(('Number of files before and',
-                                                 'after not the same.',
+                                estr = ' '.join(('Number of files before',
+                                                 nold, 'and',
+                                                 'after', nnew, 'not the same.',
                                                  'Something has gone wrong for',
```
aburrell, line 833:
```suggestion
                                estr = ' '.join(('Number of files before',
                                                 repr(nold), 'and',
                                                 'after', repr(nnew), 'not the same.',
'Something has gone wrong for',
```

Only supposed to join strings.

pysat/_orbits.py
```diff
@@ -380,10 +380,10 @@ def _calc_orbits(self):
         else:
             # Iterating by date.  We need to check step (frequency string)
             # against width (timedelta)
-            step = pds.tseries.frequencies.to_offset(self.inst._iter_step)
-            step = dt.timedelta(seconds=pds.Timedelta(step).total_seconds())
-            root = dt.datetime(2001, 1, 1)
-            if root + step < root + self.inst._iter_width:
+            istep = pds.tseries.frequencies.to_offset(self.inst._iter_step)
+            istep = dt.timedelta(seconds=int(istep.nanos * 1.0e-9))
```
aburrell, line 384: Do we have any reason to assume pandas will never change the frequency of their output?  This seems to have introduced an assumption that we have no control over and could eventually break, while the previous code made no such assumptions.

aburrell (changes requested):
Partial review.  The changes you made to the meta unit tests are basically just removing the failing tests instead of fixing the use case they were testing.  Those tests need to remain unchanged.

pysat/_orbits.py
rstoneback: I didn't change this code, this is your code. I asked you about it in your pull and you said it was necessary. You are welcome to suggest new code, if you have an improved process, or you can create an issue.

pysat/tests/test_meta.py
rstoneback: The test does use string inputs for Meta, it uses the correct strings.

pysat/utils/files.py
rstoneback: This was a debug check that I inadvertently left in. Sometimes the test failed but of course stopped failing when I put in the debug statement.

pysat/tests/test_meta.py
aburrell: No, it's using attribute labels.  It's not using various strings that a user could use instead of attributes.  I provided this as context

More importantly, you changed a unit test that was failing to make it pass instead of fixing the code so that a previously passing unit test no longer fails.  This is the core of the issue.

pysat/_orbits.py
rstoneback: I restored the original code but it fails with errors. 
```
ValueError: Value must be Timedelta, string, integer, float, timedelta or convertible, not Day
```

pysat/tests/test_meta.py
rstoneback: > No, it's using attribute labels. It's not using various strings that a user could use instead of attributes. 

I asked you on multiple occasions about that new assignment code, as well as any use cases, or tests.  You chose to not answer. This unit test looked like a test for the new meta assignment you had added. This test failed when I removed the new assign by value code. Plus all of the meta names matched the wrong aspect of labels.

The docstring says it is testing equality between Meta instances, not anything else. 

> I provided this as context
> 

I don't see anything, I think something got lost.

> More importantly, you changed a unit test that was failing to make it pass instead of fixing the code so that a previously passing unit test no longer fails. This is the core of the issue.

Per this standard I removed your changes to unit tests.

pysat/_orbits.py
rstoneback: The original code fails with errors. 
```
ValueError: Value must be Timedelta, string, integer, float, timedelta or convertible, not Day
```

pysat/_meta.py
```diff
@@ -454,11 +454,26 @@ def __setitem__(self, data_vars, input_dat):
                         # Update the MetaLabels object and the existing
                         # metadata to ensure all data have all labels
                         self.labels.update(iattr, ikey, itype)
```
rstoneback, line 477: This is where the problem comes up. If users use a metadata name like 'name' then the labels.update recognizes it matches a labels attribute and then updates the key from 'name' to 'long_name'. Simply disabling this call leads to other errors. In progress.

rstoneback: Technically this test has never passed correctly. While the test did pass it is because pandas filled in the gaps. Meta currently allows people to assign new metadata names, like 'name', even though this assignment conflicts with an existing labels attribute 'name'. When a user tries to assign a new metadata value then pysat tries to make this a label attribute, which conflicts with the existing attribute. This process should generate an error instead. Users can assign both the metadata key and value with the labels class, so users can have both 'name' and 'long_name' if the follow that path.

Anyway, here is some debugging to demonstrate this issue:
REMOVED DEBUGGING IN EDIT

rstoneback: UPDATE: I RAN THIS EXAMPLE WITH CODE THAT WAS TOO NEW. MY ORIGINAL CODE PLAN WAS CORRECT. THIS IS WHAT I GET FROM ORIGINAL CODE
```
In [4]: print(cmeta.data)
          units long_name       notes              desc value_min value_max fill           name min_val max_val fill_val
test_var  testU  test_var  test notes  test description       NaN       NaN  NaN  test variable     0.0    10.0     -1.0
```

THE PREVIOUS INFORMATION WAS INCORRECT

pysat/tests/test_meta.py
```diff
@@ -504,20 +512,23 @@ def test_value_inequality(self, label_key):
...
+                     "desc": "test description",
                      "min_val": 0.0,
                      "max_val": 10.0,
-                     "fill_val": -1.0}
+                     "fill_val": -1.0,
+                     "value_min": -999.0}
 
         self.meta['test_var'] = meta_dict
 
-        if isinstance(meta_dict[label_key], str):
```
rstoneback, line 517: Removing my changes and aburrells changes to test as it used to be.

rstoneback:
Tests are failing for an unrelated issue: https://github.com/coverallsapp/github-action/issues/264

pysat/_meta.py
rstoneback: Our recent pysat developer meeting revealed the misunderstanding. I'm not removing the feature to use custom string values in metadata. I am working to ensure that feature works robustly. There are a few attributes that are defined in the meta.labels class, like 'name', 'units', 'desc', 'fill_val' etc. Trying to assign metadata under the same string causes a name conflict that leaves the meta class unable to function 100%. In the default labels case, the only strings that will cause a problem are 'name', 'fill_val', 'min_val', or 'max_val'. In general though the other attributes can come into play, 'units', 'notes', 'desc'. What I'm doing is making sure we don't have a name conflict by dealing with the 7 'no no' words. All the other custom strings are supported.

rstoneback:
I haven't forgotten about this pull. Unfortunately I've been dealing with a back issue. The good news is I should be back on things next week.

pysat/_instrument.py
```diff
@@ -883,17 +884,32 @@ def __getitem_xarray__(self, key, data=None):
 
         if isinstance(key, tuple):
             if len(key) == 2:
-                # Support slicing time, variable name
+                # Support slicing, first extract just the desired variable(s)
                 if isinstance(key[1], slice):
```
aburrell, line 890: I'd prefer having the extra comments that were deleted, the still appear to be correct and there's no reason to reduce clarity.

pysat/_instrument.py
```diff
@@ -883,17 +884,32 @@ def __getitem_xarray__(self, key, data=None):
...
+                #     # `key[0]` must be linked to the epoch.
+                #     key_dict = {'indexers': {epoch_name: key[0]
+                #                              for epoch_name in epoch_names}}
+                # elif len(data_subset.dims) == 1:
+                #     # `key[0]` must be linked to the coordinate dimension
+                #     key_dict = {'indexers': {dname: key[0]
+                #                              for dname in data_subset.dims}}
+                # else:
+                #     raise KeyError('Unable to slice data as requested')
+
```
aburrell: Why is this commented code block here?  If it's for future reference an issue would be more appropriate.

pysat/_instrument.py
```diff
@@ -1073,20 +1100,51 @@ def __setitem__(self, key, new_data):
...
+                    #     if self.data[var_key].shape == key[0].shape and len(
+                    #             ind_keys) == 1 and type(key[0].dtype) in [
+                    #                 bool, str_type]:
+                    #         # This is a mask, using where does the opposite of
+                    #         # what is expected by assigning a mask so do the
+                    #         # inverse
+                    #         self.data[var_key] = self.data[var_key].where(
+                    #             ~key[0], in_data)
+                    #     else:
+                    #         raise ierr
```
aburrell: If you don't want to put all of this in an issue, put it in a different branch.

pysat/_instrument.py
```diff
@@ -1848,21 +1906,34 @@ def _get_var_type_code(self, coltype):
...
                      np.int8: 'i1', np.uint64: 'u8', np.uint32: 'u4',
                      np.uint16: 'u2', np.uint8: 'u1', np.float64: 'f8',
                      np.float32: 'f4', np.datetime64: 'i8'}
 
+        try:
+            str_types = [str, np.str_, np.bytes_, np.dtypes.StrDType,
+                         np.dtypes.StringDType, np.dtypes.BytesDType,
+                         pds.StringDtype]
+        except AttributeError:
+            # TODO(#1227) np.dtypes not introduced until somewhere around np1.25
```
aburrell:
```suggestion
            # TODO(#1227) np.dtypes not introduced until somewhere around numpy version 1.25
```

aburrell (commented):
Almost there.  Biggest outstanding issue is the commented code blocks that need to live someplace else.

aburrell (approved):
I am still not happy about removing the comments for no reason, but I'm not going to hold up the PR over it.

pysat/_instrument.py
rstoneback: I updated these just to move forward but you made these changes to the comments.

rstoneback:
> I am still not happy about removing the comments for no reason, but I'm not going to hold up the PR over it.

 First, you deleted the comments. Second, I added them back.

pysat/_orbits.py
rstoneback: I also noticed that no issue was created for this despite being marked as resolved.

